PR #10, opened 2019-01-24
Add fish shell setup to `env` command and README

elliottsj:
I've been testing fnm in fish shell, and it works great! Just this one setup step is needed to get it working.

UPDATE: introduced a `--fish` flag to the env command, so fish users can add this line to their `config.fish` to set up fnm:

```fish
eval (fnm env --fish)
```

---

PREVIOUS DESCRIPTION:

The drawback of this is that the `$HOME/.fnm/current/bin` implementation detail is not hidden behind the `fnm env` command.

Another possible solution is to modify the `env` command to output the fish setup commands. e.g.

```
$ fnm env --fish
set PATH $HOME/.fnm/current/bin $PATH
```

Then the README can say `eval (fnm env --fish)`. I'm willing to contribute this; just need some time to get my environment set up and learn a bit of ReasonML 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,12 @@
   ```bash
   eval `fnm env`
   ```
+  
+  If you are using [fish shell](https://fishshell.com/), add this line to your `config.fish` file:
+  
+  ```fish
+  set PATH $HOME/.fnm/current/bin $PATH
+  ```
 
 ## Future Plans
 - [ ] Add a simpler way of installing it (`curl | bash`?)
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
   If you are using [fish shell](https://fishshell.com/), add this line to your `config.fish` file:
   
   ```fish
-  set PATH $HOME/.fnm/current/bin $PATH
+  eval (fnm env --fish)
   ```
 
 ## Future Plans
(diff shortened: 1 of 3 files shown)


Schniz:
Oh, good catch! I haven’t used fish for a long time but I remember how hard it was to setup things with all the bash scripts, so thank you for that!

I think I’d rather have the `—fish` option, and also we can make a feature test with a fish shell that tests it on CI, making sure we don’t ruin the integration. Making every configuration necessary inside the binary itself can help us go through any change we’d like. What do you think?

(We can also infer the shell type from the process tree but it seems very error prone and hard, maybe an `—auto` option!)

elliottsj:
I agree! Just pushed a change adding a `--fish` flag to the env command.

The feature test + CI integration sounds like a great idea. Do you have a suggested way to set up such a test? Based on the current structure, I can try adding a `feature_tests/fish/run.sh` bash script which will install fish shell and run something like 

```bash
fish ./test.fish
```

where `test.fish` is a fish script which does something similar to the current `feature_tests/basic/run.sh`.

Not sure if it's a good idea to install fish as part of the test though. Probably better to download & run fish without affecting the system it's running on. Perhaps via Docker? I couldn't find an "official-looking" fish shell container on Docker Hub, but here's an example: https://hub.docker.com/r/frapsoft/fish/dockerfile

Issue #5, opened 2019-01-24
Release a Homebrew formula


yardnsm:
It'll be awesome to have a [Homebrew formula](https://docs.brew.sh/Formula-Cookbook) so it can be installed using Homebrew. I can tackle this if you'd like.

Awesome work btw!

Schniz:
I would be happy to get help with that!

flesch:
:wave: Hi @Schniz and @yardnsm!

I just submitted a pull request adding this to Homebrew: https://github.com/Homebrew/homebrew-core/pull/36744.

Schniz:
Yay! Thanks @flesch 😸

Schniz:
Let's hope they'll merge it. If they do, we can add `brew install fnm` to the README 😄

Schniz:
`brew install Schniz/tap/fnm` 😄

PR #14, opened 2019-01-29
install detail

gozeon:


Schniz:
Hi @gozeon! thanks for the contribution 🎉 
Great work. I think that if we'll have a script - we don't need to explain how to install it, people could just install it using the script.

What do you think about adding a script, like `install.sh` on the root of the repo, so potential users could just curl it?

Eventually I'd want to add a homebrew formula somehow (#5), so it'd be easier to install on Macs

README.md
```diff
@@ -31,6 +31,18 @@
   eval (fnm env --fish)
   ```
 
+### install reference
+
+```bash
+cd ~
+wget https://github.com/Schniz/fnm/releases/download/v1.1.0/fnm-linux.zip
```
Schniz: Maybe we should always download the latest version? here's a gist that helps with that: https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c#file-get_latest_release-sh-L2-L6

```bash
get_latest_release() {
  curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
    grep '"tag_name":' |                                            # Get tag line
    sed -E 's/.*"([^"]+)".*/\1/'                                    # Pluck JSON value
}

LATEST_RELEASE=$(get_latest_release Schniz/fnm)
URL=https://github.com/Schniz/fnm/releases/download/$LATEST_RELEASE/fnm-linux.zip
```

Also, do you think we should have one script for both Mac and Linux?

satvikpendem:
Yes, a script would be very nice. Currently I downloaded and unzipped the zip to ~/.fnm, and added this to $PATH, but when I do ```eval `(fnm env)` ```, the command is not found.

Schniz:
Hey @satvikpendem, sorry to hear! can you share your script please? 😄 maybe we can help 😸

satvikpendem:
I don't have any script, I just followed the instructions on the README. Could you go into more detail about how exactly to download and run the program? I'm on linux command line by the way.

README.md
```diff
@@ -31,6 +31,18 @@
...
+unzip fnm-linux.zip
+chmod +x ~/fnm-linux/fnm-linux
+ln ~/fnm-linux/fnm-linux /usr/local/bin/fnm
+echo "eval `fnm env`" >> ~/.bashrc && source ~/.bashrc
+enjoy it :)
```
itaisteinherz: It should be clear that this is only a comment and not a command to run (a bit of a nitpick but still):

```suggestion
$ # enjoy it :)
```

Also, there should be added `$` prompts before the commands above to clarify that those are the commands users need to run in order to install `fnm`.

Schniz:
Hi! I'm closing it because we have a working installation script. Thanks for your contribution though, it made the script happen! ✨

PR #17, opened 2019-01-30
Use xz files instead of gz

### LINKED ISSUE #7: Download Node versions with `.tar.xz` files instead of `.tar.gz`
`.tar.xz` formats tend to be smaller than `.tar.gz` files, for example on version [v11.7.0](http://nodejs.org/dist/v11.7.0/):

* `node-v11.7.0-darwin-x64.tar.gz`: 16947291
* `node-v11.7.0-darwin-x64.tar.xz`: 11496916

more than 5 megabytes of difference. This could make installation even faster.

Schniz:
`.tar.xz` files are usually smaller than `.tar.gz`, so it'll be faster to download those.

Fixes #7

### DIFF
--- a/executable/Install.re
+++ b/executable/Install.re
@@ -44,7 +44,10 @@ let main = (~version as versionName) => {
   let%lwt filepath =
     Versions.getFileToDownload(~version=versionName, ~os, ~arch);
   let tarDestination =
-    Filename.concat(Directories.downloads, versionName ++ ".tar.gz");
+    Filename.concat(
+      Directories.downloads,
+      versionName ++ Versions.Remote.downloadFileSuffix,
+    );
 
   Console.log(
     <Pastel>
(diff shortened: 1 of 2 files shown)


lzhang:
Unfortunately this seems to have broken the ability to install certain versions. For example:

https://nodejs.org/dist/v0.10.29/

ljharb:
Indeed, older versions don't have xz builds available.

Schniz:
> Unfortunately this seems to have broken the ability to install certain versions. For example:
> 
> [nodejs.org/dist/v0.10.29](https://nodejs.org/dist/v0.10.29/)

The question is - whether should we support a Node version that isn't supported for more than 4 years. We can add support for `.gz`, but I don't think it is worth it. We might need to have better errors though.

lzhang:
In my view, the utility of a tool like fnm lies precisely in the ability to switch versions on legacy projects that are using extremely old versions of node. Narrowing version compatibility to marginally speed up decompression seem like a bad tradeoff. That being said, I understand the viewpoint that legacy version support isn't worth the trouble.

Full compatibility goes beyond tar.gz support. I was looking into this a little closer, and I noticed that the directory structure changes over time. Also sometimes only `.msi` files are available for Windows.

PR #27, opened 2019-02-11
skip installation if the version is already installed

kentac55:
~1. When install specify version from `fnm ls-remote`, the version which has been installed contains ANSI codes and `fnm install` raise internal error.~

i reverted it.

```
$ fnm install $(fnm ls-remote | grep -E "\* v[0-9]?[0,2,4,6,8]" | tail -n 1 | sed -r -e "s/\* //g")

Looking for node v10.15.1 for linux x64
     (Failure hd)
     Raised at file "pervasives.ml", line 32, characters 17-33
     Called from file "list.ml" (inlined), line 27, characters 10-23
     Called from file "library/Http.re", line 36, characters 25-41
     Called from file "library/Http.re", line 43, characters 2-27
     Called from file "src/core/lwt.ml", line 1929, characters 23-26
     Re-raised at file "library/Http.re", line 41, characters 2-133
     Re-raised at file "library/Versions.re", line 80, characters 2-580
     Re-raised at file "executable/Install.re", line 45, characters 2-942
     Re-raised at file "src/core/lwt.ml", line 2987, characters 20-29
     Called from file "src/unix/lwt_main.ml", line 26, characters 8-18
     Called from file "cmdliner_term.ml", line 25, characters 19-24
     Called from file "cmdliner.ml", line 116, characters 32-39
```

2. skip to install the version which has been already installed

### DIFF
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -149,10 +149,10 @@ let getRemoteVersions = () => {
 
 let throwIfInstalled = versionName => {
   switch (getInstalledVersions()) {
+  | Error(_) => Lwt.return()
   | Ok(x) =>
     Array.to_list(x)
     |> List.exists(x => Local.(x.name == versionName))
     |> (x => x ? Lwt.fail(Already_installed(versionName)) : Lwt.return())
-  | Error(_) => Lwt.return()
   };
 };
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -152,8 +152,7 @@ let throwIfInstalled = versionName => {
   |> Result.fold(
        _ => Lwt.return(),
        xs =>
-         Array.to_list(xs)
-         |> List.exists(x => Local.(x.name == versionName))
+         Array.exists(x => Local.(x.name == versionName), xs)
          |> (
            x => x ? Lwt.fail(Already_installed(versionName)) : Lwt.return()
          ),
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -148,18 +148,11 @@ let getRemoteVersions = () => {
 };
 
 let throwIfInstalled = versionName => {
-  switch(getInstalledVersions()) {
-  | Ok(x) => {
+  switch (getInstalledVersions()) {
+  | Ok(x) =>
     Array.to_list(x)
-    |> List.exists(x => {
-      open Local;
-      x.name == versionName;
-    })
-    |> x => switch x {
-    | true => Lwt.fail(Already_installed(versionName));
-    | false => Lwt.return();
-    };
-  };
-  | Error(_) => Lwt.return();
+    |> List.exists(x => Local.(x.name == versionName))
+    |> (x => x ? Lwt.fail(Already_installed(versionName)) : Lwt.return())
+  | Error(_) => Lwt.return()
   };
 };
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -59,8 +59,7 @@ let format = version => {
     | exception _ => version
     };
 
-  let ansiCharRegex = Str.regexp({|\\027\[[0-9]+m|});
-  Str.global_replace(ansiCharRegex, "", String.escaped(version));
+  version;
 };
 
 let endsWith = (~suffix, str) => {
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -58,7 +58,8 @@ let format = version => {
     | exception _ => version
     };
 
-  version;
+  let ansiCharRegex = Str.regexp({|\\027\[[0-9]+m|});
+  Str.global_replace(ansiCharRegex, "", String.escaped(version));
 };
 
 let endsWith = (~suffix, str) => {
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -148,11 +148,14 @@ let getRemoteVersions = () => {
 };
 
 let throwIfInstalled = versionName => {
-  switch (getInstalledVersions()) {
-  | Error(_) => Lwt.return()
-  | Ok(x) =>
-    Array.to_list(x)
-    |> List.exists(x => Local.(x.name == versionName))
-    |> (x => x ? Lwt.fail(Already_installed(versionName)) : Lwt.return())
-  };
+  getInstalledVersions()
+  |> Result.fold(
+       _ => Lwt.return(),
+       xs =>
+         Array.to_list(xs)
+         |> List.exists(x => Local.(x.name == versionName))
+         |> (
+           x => x ? Lwt.fail(Already_installed(versionName)) : Lwt.return()
+         ),
+     );
 };
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -147,7 +147,7 @@ let getRemoteVersions = () => {
   |> Lwt.return;
 };
 
-let isAlreadyInstalled = versionName => {
+let throwIfInstalled = versionName => {
   switch(getInstalledVersions()) {
   | Ok(x) => {
     Array.to_list(x)
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -8,6 +8,7 @@ module Local = {
 };
 
 exception Version_not_found(string);
+exception Already_installed(string);
 
 module Remote = {
   type t = {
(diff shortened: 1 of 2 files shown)


library/Versions.re
```diff
@@ -145,3 +147,20 @@ let getRemoteVersions = () => {
      )
   |> Lwt.return;
 };
+
+let isAlreadyInstalled = versionName => {
```
Schniz: I would assume it'll return a `bool`, whether the version is already installed. What do you think? Another thing we can do is "throwIfInstalled" or something like that.

Schniz:
Thanks @kentac55! it looks much better and I hope #28 will help us with passing results between calls.

Can you please rename the function so it would be `throwIfMissing` or something that would match its type signature of `Lwt.t(unit)`?

kentac55:
@Schniz Thank you for your reviewing! And sorry for my confusing(it's my first time to write reasonml(and ocaml) and I misread your review :dizzy_face: )

library/Versions.re
```diff
@@ -145,3 +146,20 @@ let getRemoteVersions = () => {
      )
   |> Lwt.return;
 };
+
+let throwIfInstalled = versionName => {
+  switch(getInstalledVersions()) {
+  | Ok(x) => {
+    Array.to_list(x)
+    |> List.exists(x => {
```
Schniz: You can use `Array.exists` too, so no conversion is needed

library/Versions.re
```diff
@@ -145,3 +146,20 @@ let getRemoteVersions = () => {
...
+      open Local;
```
Schniz: opening `Local` here feels weird: maybe we can move this function to the `Local` module? I mean, _it is_ something related to local versions only

`Versions.Local.throwIfInstalled` sounds reasonable to me - and no `open` will be needed!

library/Versions.re
```diff
@@ -145,3 +146,20 @@ let getRemoteVersions = () => {
...
+      x.name == versionName;
+    })
+    |> x => switch x {
```
Schniz: There's a command: `esy fmt` that runs `refmt` on the codebase. Once you'll run it, you see that `switch` over boolean will be converted to a ternary. It'll be prettier 😄 please run it against your code before pushing (I'm adding it as a commit hook on #30)

library/Versions.re
```diff
@@ -145,3 +146,20 @@ let getRemoteVersions = () => {
...
+    | true => Lwt.fail(Already_installed(versionName));
+    | false => Lwt.return();
+    };
+  };
+  | Error(_) => Lwt.return();
```
Schniz: I think it'll be more readable if the short paths will be on the top of the switch cases, so they would feel like early returns:

```reason
switch (x) {
  | Error(_) => Lwt.return()
  | Ok(o) => {
       ...
  }
}
```

library/Versions.re
```diff
@@ -145,3 +146,20 @@ let getRemoteVersions = () => {
      )
   |> Lwt.return;
 };
+
+let throwIfInstalled = versionName => {
+  switch(getInstalledVersions()) {
```
Schniz: Instead of switching, you can also fold it to a value - an error is like an empty array in this case:

```reason
getInstalledVersions() |> Result.fold(_ => [||], xs => xs)
```

After folding, the code will probably be much cleaner 😄

Schniz: Folding is weird for newcomers, I hope that I got it right too. What it means:

```reason
let fold = (error, ok, result) =>
  switch (result) {
  | Error(x) => error(x)
  | Ok(x) => ok(x)
  }
```

It helps us take the wrapped value (the array) and the error, and convert it into one data structure. In this case, we don't care about the errors - we can assume it's an empty array, so folding it to an array makes sense, and we even remove one level of the pattern matching!

library/Versions.re
kentac55: yes...I know `Array.exists` but I rewrote with it like below, compiler said type error. :cry: 

How should I fix it?
```ocaml
    Array.exists(x => {
      open Local;
      x.name == versionName;
    })
    |> x => switch x {
    | true => Lwt.fail(Already_installed(versionName));
    | false => Lwt.return();
    };
```
```
 ERROR  library/Versions.re:158 4-10

155 ┆     x.name == versionName;
156 ┆   })
157 ┆   |> x => switch x {
158 ┆   | true => Lwt.fail(Already_installed(versionName));
159 ┆   | false => Lwt.return();
160 ┆   };
161 ┆ };

This pattern type doesn't match what is expected.

  This type: bool
  Expecting: array(Local.t) => bool

```

Schniz: I'm going to need more context 😄 

What about the following:

```reason
Array.exists(x => {
  open Local;
  x.name == versionName
}, x)

// or

x |> Array.exists(x => {
  open Local;
  x.name == versionName
})
```

Usually, when the compiler starts throwing stuff at me that I don't understand, I'm starting to break down my code to "compilable" pieces: trying to break down things and extract variables. Using `x` as names and piping stuff to anonymous functions tend to be confusing so I'd rather avoid it unless it is a one-liner or something really general

kentac55: Thank you! I understood it! :heart:

library/Versions.re
kentac55: I got it!

library/Versions.re
kentac55: I agree it! and I try it... but `getInstalledVersions()` belong to `Version` module, so I cannot use it in `Versions.Local`. I Tried to write  `Version.getInstalledVersions()` but it caused build error.
How can I use parent module's function in child? (If this resolution is in [docs](https://reasonml.github.io/docs/en/module), I'm so sorry)

Schniz: You're totally right. I think it's not something related to parent-child, but more of the order of declaration: https://reasonml.github.io/try?rrjsx=true&reason=DYUwLgBA7g9gTsAJhAvBARLBj0G4BQ+AtjIgK6gQASIwwMqEA3vhBKJAA6MAUAlKgB8ETnACWAOzAB9EBMTBJIHuhp0G6CAGot0eEj4EAvoTX0AdJ364gA

I'd say we leave it as it is right now, and later on we'll have to refactor. In my opinion, `Versions.getInstalledVersions` should be `Versions.Local.getInstalled` (makes more sense), but it shouldn't bother you now

library/Versions.re
```diff
@@ -145,3 +146,15 @@ let getRemoteVersions = () => {
      )
   |> Lwt.return;
 };
+
+let throwIfInstalled = versionName => {
+  getInstalledVersions()
+  |> Result.fold(
+       _ => Lwt.return(),
```
Schniz: If we would fold it to an empty array, we could push down the knowledge of Lwt and take out one indentation level:

```reason
let installed = getInstalledVersions()
|> Result.fold(_ => [||], xs => xs) // would return an empty array for errors
|> Array.exists(x => Local.(x.name == versionName))

if (installed) {
  Lwt.fail(Already_installed(versionName));
} else {
  Lwt.return();
}
```

Schniz: that being said, I'm really not sure we need to even care about Lwt here. We can stick to `isVersionInstalled` like you mentioned before that, and leave the logic of throwing an error to the executable itself (which also handles the error):

```reason
// library/Versions.re

let isVersionInstalled = versionName =>
  getInstalledVersions()
  |> Result.fold(_ => [||], xs => xs) // would return an empty array for errors
  |> Array.exists(x => Local.(x.name == versionName));
```

```reason
// executable/Install.re

let%lwt _ = if (Versions.isVersionInstalled(someVersion)) {
  Lwt.fail(Already_installed(someVersion));
} else {
  Lwt.return();
}
```

Schniz:
Hey @kentac55, what's up? I want to merge this and #30, and I don't want to create conflicts for your branch by merging the other 😄

theandrewlane:
@Schniz/@kentac55 pleeaaaseee get this merged in soon ♥️

Schniz:
Now that I think about it, in #30 I'm making `readdir` async (with Lwt), therefore this function would be async as well. I'm merging 😄

Schniz:
@kentac55, thanks for all your hard work 😄 🏆 🌮

kentac55:
@Schniz Sorry for late response. I was too busy to response this PR in few days. Thanks for fix and merge it and I try to study reasonml :heart:

PR #36, opened 2019-02-14
Support Node.js mirrors

### LINKED ISSUE #32: Allow downloading versions from a Node.js mirror
This is a feature request.

When I already have `~/.fnm/downloads/v11.9.0.tar.xz` exists on disk, if I run `fnm install 11.9.0`, I would like fnm to try to unarchive it first and starting to download the fresh package only if unarchive exits with none zero status.

I use Taobao's node dist mirror. It's faster than Nodejs's 1st party one since I'm in China.(Taobao is a Alibaba company)

As a example:

```
# I use this url
https://npm.taobao.org/dist/v11.9.0/node-v11.9.0-darwin-x64.tar.xz
# instead of
https://nodejs.org/dist/v11.9.0/node-v11.9.0-darwin-x64.tar.xz
```

Schniz:
Fixes #32 

cc @haishanh 😸

### DIFF
--- a/executable/Env.re
+++ b/executable/Env.re
@@ -1,14 +1,24 @@
 open Fnm;
 
-let run = isFishShell => {
+let run = (isFishShell, nodeDistMirror) => {
   if (isFishShell) {
-    Console.log(
-      Printf.sprintf("set PATH %s/bin $PATH", Directories.currentVersion),
-    );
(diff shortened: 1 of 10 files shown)


haishanh:
Awesome 🎉

PR #42, opened 2019-02-16
Add support for .node-version files

### LINKED ISSUE #8: Add support for .node-version files
Most version managers support the `.node-version` format:

Tool Name | Config File | Plattform | Notes
-- | -- | -- | --
[nvs](https://github.com/jasongin/nvs) | .node-version | Mac / Linux / Windows |  
[nvm](https://github.com/creationix/nvm) | .nvmrc | Mac / Linux |
[nodenv](https://github.com/nodenv/nodenv) | .node-version | Mac / Linux |  
[nodist](https://github.com/nullivex/nodist) | .node-version | Windows |  
[avn](https://github.com/wbyoung/avn) | .node-version | Mac / Linux |  
[n](https://github.com/tj/n) |   | Mac / Linux | Does not seem to have a config file option

Could you please also add support?

Dean177:
I attempted to add support for `.node-version` files.

I'm brand new to the reason ecosystem so have absolutely no idea what I am doing.

The files in the `esy.lock` directory seem to have changed quite substantially, have I done something wrong to cause this?

Resolves #8


library/Dotfiles.re
```diff
@@ -0,0 +1,18 @@
...
+let versionString = fileName => {
+  Lwt_io.lines_of_file(fileName)
+  |> Lwt_stream.to_list
+  |> Lwt.map(List.hd)
+  |> Lwt.map(String.trim);
+};
+
+let getVersion = () => {
+  let%lwt cwd = Lwt_unix.getcwd();
+  try%lwt (versionString(Filename.concat(cwd, ".node-version"))) {
```
Schniz: What if `versionString` would catch the errors and provide `option(string)`, that means that `Some("file contents")` or `None` when file not found?
then, this code would be easier to follow and Reason's amazing pattern matching will help us:

```reason
let%lwt nodeVersion = Filename.concat(cwd, ".node-version") |> versionString
and nvmrc = Filename.concat(cwd, ".nvmrc") |> versionString;

switch (nodeVersion, nvmrc) {
| (None, None) => Lwt.fail(Version_Not_Provided)
| (Some(_), Some(_)) => Lwt.failwith("You have both .node-version and .nvmrc!")
| (Some(version), None)
| (None, Some(version)) => Lwt.return(version)
};
```

executable/Use.re
```diff
@@ -71,11 +71,11 @@ let run = (~version, ~quiet) =>
       </Pastel>,
     )
     |> Lwt.return
-  | Nvmrc.Version_Not_Provided =>
+  | Dotfiles.Version_Not_Provided =>
     log(
       ~quiet,
       <Pastel color=Pastel.Red>
-        "No .nvmrc was found in the current directory. Please provide a version number."
+        "No .nvmrc or .node-version file was found in the current directory. Please provide a version number."
```
Schniz: 👏

library/Dotfiles.re
Schniz: or even:

```reason
let%lwt nodeVersion = Filename.concat(cwd, ".node-version") |> versionString
and nvmrc = Filename.concat(cwd, ".nvmrc") |> versionString;

switch (nodeVersion, nvmrc) {
| (None, None) => Lwt.fail(Version_Not_Provided)
| (Some(version), Some(_))
| (Some(version), None)
| (None, Some(version)) => Lwt.return(version)
};
```

Schniz:
> I'm brand new to the reason ecosystem so have absolutely no idea what I am doing.

May I just add, that it sure doesn't look this way.
Looks very very good 😄

library/Dotfiles.re
Dean177: Good idea and thanks for the feedback. I think in the case where a user has both a `.node-version` and an `.nvmrc` we can be permissive if they contain matching versions (since they might just want to support multiple clients).

Whats the difference between:
```
let%lwt nodeVersion = Filename.concat(cwd, ".node-version") |> versionString
and nvmrc = Filename.concat(cwd, ".nvmrc") |> versionString;
```
and 
```
let%lwt nodeVersion = Filename.concat(cwd, ".node-version") |> versionString;
let%lwt nvmrc = Filename.concat(cwd, ".nvmrc") |> versionString;
```
The docs mention parallel binding: https://ocsigen.org/lwt/3.2.1/api/Ppx_lwt
Does this mean they will actually be run in parallel?

Schniz: Indeed, it’s like `Promise.all`!

library/Dotfiles.re
```diff
@@ -0,0 +1,33 @@
...
+
+let getVersion = () => {
+  let%lwt cwd = Lwt_unix.getcwd();
+  let%lwt nodeVersion = versionString(Filename.concat(cwd, ".node-version"))
+  and nvmrc = versionString(Filename.concat(cwd, ".nvmrc"));
+
+  switch (nodeVersion, nvmrc) {
+  | (None, None) => Lwt.fail(Version_Not_Provided)
+  | (Some(nodeVersionString), Some(nvmrcString)) =>
+    if (nodeVersionString === nvmrcString) {
```
Schniz: This can even be reduced further using:

```reason 
  | (Some(v1), Some(v2) when v1 == v2 =>
```
Then, no nesting in our pattern matching! Wdyt?

Schniz: Or even better; check if it isn’t equal first :)

PR #48, opened 2019-02-18
Parameterize the install script for relative install of fnm

from-nibly:
What are your thoughts on parameterizing the install script for local install? I know this seems weird but it means that I don't have to rely on developers having any tools already installed except bash and some way to edit the files. Here is an actual example script setup I would use.

Here is an example local install script `installLocal.sh`
```bash
#!/usr/bin/env bash

DESIRED_NODE_VERSION=v$(cat package.json | grep -Ei "\"node\"" | grep -Eoi "[0-9]+\.[0-9]+\.[0-9]+") 

export FNM_DIR="$(pwd)/.fnm"
if [ ! -d "$FNM_DIR" ]; then
    curl 'https://raw.githubusercontent.com/Schniz/fnm/master/.ci/install.sh' | INSTALL_DIR=`pwd` bash
fi

set -e

NODE_DIR="$FNM_DIR/node-versions/$DESIRED_NODE_VERSION/installation/bin"
export PATH=$FNM_DIR:$NODE_DIR:$PATH

#TODO: support fish
eval `fnm env --multi`

if [ ! -f "$NODE_DIR/node" ]; then
    fnm install $DESIRED_NODE_VERSION
fi

sumOfPackageJson() {
    cat package.json | md5sum | cut -d' ' -f1
}

if [ ! -f ".packageRevision" ] || [ "$(cat .packageRevision)" != "$(sumOfPackageJson)" ]; then
    sumOfPackageJson > .packageRevision
    npm install
fi
```

then in script called `nodew`

```bash
#!/usr/bin/env bash
cd "$( dirname "${BASH_SOURCE[0]}" )"
source ./installLocal.sh
node "$@"
```

and `npmw`
```bash
#!/usr/bin/env bash
cd "$( dirname "${BASH_SOURCE[0]}" )"
source ./installLocal.sh
npm "$@"
```

and for completeness sake I usually go one step further and do an `appw`
```bash
#!/usr/bin/env bash
cd "$( dirname "${BASH_SOURCE[0]}" )"
source ./installLocal.sh
node ./app/index.js "$@"
```

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -30,6 +30,10 @@ parse_args() {
 	  SKIP_SHELL="true"
 	  shift # past argument
 	  ;;
+	  *)
+	  echo "Unrecognized argument $key"
+	  exit 1
+	  ;;
   esac
   done
 }
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -12,6 +12,31 @@ else
   exit 1
 fi
 
+INSTALL_DIR="$HOME/.fnm"
+
+# Parse Flags
+parse_args() {
+  while [[ $# -gt 0 ]]
+  do
+  key="$1"
+
+  case $key in
+	  -d|--install-dir)
+	  INSTALL_DIR="$2"
+	  shift # past argument
+	  shift # past value
+	  ;;
+	  -s|--skip-shell)
+	  SKIP_SHELL="true"
+	  shift # past argument
+	  ;;
+  esac
+  done
+}
+
+parse_args "$@"
+
+
 get_latest_release() {
   # Taken from https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
   curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
@@ -26,11 +51,11 @@ download_fnm() {
 
   echo "Downloading $URL..."
 
-  mkdir -p $HOME/.fnm &> /dev/null
+  mkdir -p $INSTALL_DIR &> /dev/null
   curl --progress-bar -L $URL -o $DOWNLOAD_DIR/$FILENAME.zip
   unzip -q $DOWNLOAD_DIR/$FILENAME.zip -d $DOWNLOAD_DIR
-  mv $DOWNLOAD_DIR/$FILENAME/fnm $HOME/.fnm/fnm
-  chmod u+x $HOME/.fnm/fnm
+  mv $DOWNLOAD_DIR/$FILENAME/fnm $INSTALL_DIR/fnm
+  chmod u+x $INSTALL_DIR/fnm
 }
 
 check_dependencies() {
@@ -112,4 +137,6 @@ setup_shell() {
 
 check_dependencies
 download_fnm
-setup_shell
+if [ "$SKIP_SHELL" != "true" ]; then 
+  setup_shell
+fi
--- a/README.md
+++ b/README.md
@@ -28,6 +28,21 @@ For `bash`, `zsh` and `fish` shells, there's an [automatic installation script](
 curl https://raw.githubusercontent.com/Schniz/fnm/master/.ci/install.sh | bash
 ```
 
+#### Parameters
+
+`--install-dir`
+
+Set a custom directory for fnm to be installed. The default is `$HOME/.fnm`.
+
+`--skip-shell`
+
+Skip appending shell specific loader to shell config file, based on the current user shell, defined in `$SHELL`. e.g. for Bash, `$HOME/.bashrc`. `$HOME/.zshrc` for Zsh. For Fish - `$HOME/.config/fish/config.fish`
+
+Example: 
+```bash
+curl https://raw.githubusercontent.com/Schniz/fnm/master/.ci/install.sh | bash --install-dir "./.fnm" --skip-shell
+```
+
 ### Manually
 
 - Download the [latest release binary](https://github.com/Schniz/fnm/releases) for your system


.ci/install.sh
```diff
@@ -12,6 +12,8 @@ else
   exit 1
 fi
 
+INSTALL_DIR=${INSTALL_DIR:-$HOME}
```
Schniz: Installing to a subdirectory of the provided INSTALL_DIR sounds counter-intuitive to me. I think I'd rather have `INSTALL_DIR=$(pwd)/.fnm (curl ... | bash)` than the other way around. What do you think?

Schniz: Also, I would highly reconsider the usage of environment variables here.

I can suggest using options, [like RVM](https://rvm.io/rvm/install):

```bash
curl ... | bash                              # regular installation
curl ... | bash -s --install-dir $(pwd)/.fnm # install to specific directory
```

.ci/install.sh
```diff
@@ -112,4 +114,6 @@ setup_shell() {
 
 check_dependencies
 download_fnm
-setup_shell
+if [ "$INSTALL_DIR" == "$HOME" ]; then
```
Schniz: I think this is cool heuristic but I'd rather have `$SKIP_SHELL_INSTALLATION` or a `--skip-shell` flag, making it more explicit. What if you don't want it to be on your `$HOME/.fnm` but `$HOME/.my-favorite-manager`?

Schniz:
And if we'll take a look at your script, you can drop some lines:

```bash
NODE_DIR="$FNM_DIR/node-versions/$DESIRED_NODE_VERSION/installation/bin"
export PATH=$FNM_DIR:$NODE_DIR:$PATH
```

These lines are already taken care of with `eval $(fnm env --multi)`, when you have the `FNM_DIR` variable exported.

I would also think that we need to have some kind of `fnm inspect` or `fnm info` to provide some data about the internals, like `fnm info $DESIRED_NODE_VERSION path`, that would return the long `$FNM_DIR/node-versions/$DESIRED_NODE_VERSION/installation`. I tend to believe that directory structure is considered private, and we can expose data with calls this way

.ci/install.sh
from-nibly: I'm good with either. But it would complicate the install script.  Just let me know which you would rather have and I can make the change.

.ci/install.sh
from-nibly: That is a good point. I can add this as well. Should I also make the name of the folder (.fnm) be changable? or should I make the install dir parameter be the entire install dir and not append `.fnm` for the user?

Schniz: I think we shouldn't append `.fnm` for the user

Schniz: The user wants to configure the install dir - so we have no assumptions there! 😜

from-nibly:
Oh good to know.

so then those two lines would become
```bash
export PATH=$FNM_DIR:$PATH
```
Because calling
```bash
eval fnm env --multi
```
 requires that fnm is already on the path.

or we could change that line to 
```bash
eval $FNM_DIR/fnm env --multi
```
Or does the install script call this for you already?

.ci/install.sh
Schniz: It does complicate things a bit, but can still be manageable with proper writing (I dislike long bash scripts for that particular reason). Maybe something like

```bash
# at top, all the defaults

INSTALL_DIR=$HOME/.fnm

parse_args() {
  while [[ $# -gt 0 ]]; do
    key=$1
    case $key in
      --install-dir)
        $INSTALL_DIR=$2
        shift
        shift
      ;;
    esac
  done
}
```

kinda readable 🤷‍♂️ what do you think?

Schniz:
> Oh good to know.
> 
> so then those two lines would become
> 
> ```shell
> export PATH=$FNM_DIR:$PATH
> ```
> 
> Because calling
> 
> ```shell
> eval fnm env --multi
> ```
> 
> requires that fnm is already on the path.
> 
> or we could change that line to
> 
> ```shell
> eval $FNM_DIR/fnm env --multi
> ```
> 
> Or does the install script call this for you already?

Nope, the install script runs on a separate bash process (not sourcing the file), so it ask the user to source it itself:

https://github.com/Schniz/fnm/blob/5442d51cd12ecb8575d766e36c6a8e4c29e15239/.ci/install.sh#L108-L110

from-nibly:
Gotcha. So I would just need to add the path for fnm or reference the binary directly when calling it.

from-nibly:
@Schniz Just made some changes to the install script. What are your thoughts?

from-nibly:
the revised installLocal.sh script would be 

```bash
#!/usr/bin/env bash

DESIRED_NODE_VERSION=v$(cat package.json | grep -Ei "\"node\"" | grep -Eoi "[0-9]+\.[0-9]+\.[0-9]+") 

export FNM_DIR="$(pwd)/.fnm"
if [ ! -f "$FNM_DIR/fnm" ]; then
    curl 'https://raw.githubusercontent.com/Schniz/fnm/master/.ci/install.sh' | bash --install-dir $FNM_DIR --skip-shell
fi

set -e

NODE_DIR="$FNM_DIR/node-versions/$DESIRED_NODE_VERSION/installation/bin"
export PATH=$FNM_DIR:$PATH
#TODO: support fish

eval `fnm env --multi`


if [ ! -f "$NODE_DIR/node" ]; then
    fnm install $DESIRED_NODE_VERSION
fi


sumOfPackageJson() {
    cat package.json | md5sum | cut -d' ' -f1
}

if [ ! -f ".packageRevision" ] || [ "$(cat .packageRevision)" != "$(sumOfPackageJson)" ]; then
    sumOfPackageJson > .packageRevision
    npm install
fi
```

.ci/install.sh
```diff
@@ -12,6 +12,28 @@ else
   exit 1
 fi
 
+# Parse Flags
+POSITIONAL=()
```
Schniz: consider adding it to a function, so it would be ordered:

```bash
# at top, all the defaults

INSTALL_DIR=$HOME/.fnm

parse_args() {
  while [[ $# -gt 0 ]]; do
    key=$1
    case $key in
      --install-dir)
        $INSTALL_DIR=$2
        shift
        shift
      ;;
    esac
  done
}

# then afterwards

parse_args $@ # (or is it `$*`?)
```

Schniz: This way the argument parsing will _override_ the default definitions (that are much more readable this way, IMO)

from-nibly: updated

.ci/install.sh
```diff
@@ -12,6 +12,33 @@ else
   exit 1
 fi
 
+INSTALL_DIR="$HOME/.fnm"
+
+# Parse Flags
+parse_args() {
+  POSITIONAL=()
```
Schniz: I can also guess we can drop the POSITIONAL arguments. No need for them here. Other than that, looks very very good.

I guess eventually you could create a binary that wraps `fnm` and does everything you script did:

```bash
w app # starts app, installs using fnm if needed
w npm # runs npm, installs using fnm if needed
w node # runs node, installs using fnm if needed
```

😄

from-nibly:
Ok there we go. Yeah I could do that but bash is good enough for now and it allows me to tweak it really easily.

.ci/install.sh
```diff
@@ -12,6 +12,31 @@ else
...
+	  -d|--install-dir)
+	  INSTALL_DIR="$2"
+	  shift # past argument
+	  shift # past value
+	  ;;
+	  -s|--skip-shell)
+	  SKIP_SHELL="true"
+	  shift # past argument
+	  ;;
+  esac
```
Schniz: oh oh I just figured out -- we might need to have the following case:

```bash
    *)
    echo "Wrong argument provided: $key"
    exit 1
    ;;
```

so we won't have infinite loops on providing the wrong argument

from-nibly: Oh dang. Good catch! No keep the comments coming. This is your project and I want to make sure I keep inline with your vision. I will go ahead and add some docs to the readme and add the breakout for unrecognized arguments.

from-nibly:
New commit pushed. Let me know what you think of the documentation.

README.md
```diff
@@ -28,6 +28,21 @@ For `bash`, `zsh` and `fish` shells, there's an [automatic installation script](
...
 
+#### Parameters
+
+`--install-dir`
+
+Set a custom directory for fnm to be installed. The default is `$home/.fnm`.
+
+`--skip-shell`
+
+Skip appending shell specific loader to shell config file e.g. `.bashrc` `.zshrc` `config.fish`
```
Schniz:
```suggestion
Skip appending shell specific loader to shell config file, based on the current user shell, defined in `$SHELL`. e.g. for Bash, `$HOME/.bashrc`. `$HOME/.zshrc` for Zsh. For Fish - `$HOME/.config/fish/config.fish`
```

README.md
```diff
@@ -28,6 +28,21 @@ For `bash`, `zsh` and `fish` shells, there's an [automatic installation script](
 curl https://raw.githubusercontent.com/Schniz/fnm/master/.ci/install.sh | bash
 ```
 
+#### Parameters
+
+`--install-dir`
+
+Set a custom directory for fnm to be installed. The default is `$home/.fnm`.
```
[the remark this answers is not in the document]
from-nibly: *face palm* not my best work lol

PR #56, opened 2019-02-19
Correct status code on `install` failures

### LINKED ISSUE #53: When a version does not exist, fnm exists with status code 0
After using `fnm install` I got an informative message `Version v8 not found!` It said that the task was not sucessfull, but It would be better if the color would be red.

I also expected the status code for unsuccessful oparation to be `1` instead of `0`.

image

ranyitz:
Fixes #53 

Change the status code to 1 for cases when the version was not found or there is no download available.


Schniz:
Thanks! 👏 
we can also make the text red, like you said 😄

PR #58, opened 2019-02-20
Adding check for OSX during writing for bash shell

### LINKED ISSUE #40: Mac OS command not found
Mac OS, if I'm not wrong, by default running .bash_profile before opening new terminal window, so when I installed fnm via curl i need to copy those lines from .bashrc to .bash_profile: 
`export PATH=$HOME/.fnm:$PATH
eval fnm env --multi`

don't know if it's really a problem but it took me little digging why fnm command not found..

maxknee:
OS X users experience issues due to .bashrc not being
sourced on start up.

Fixes #40


maxknee:
This should solve this issue: https://github.com/Schniz/fnm/issues/40#issue-410811466

PR #68, opened 2019-02-25
Infer shells automatically, and `use` versions based on the current working directory (optional)

### LINKED ISSUE #20: Infer shell type in `fnm env`
We can use `$SHELL` to see if we have `fish` or anything else, which we assume is bash compliant for now 😸:

```bash
$ echo $SHELL
/bin/zsh
```

Schniz:
Add the ability to `use` on every pwd change (bash, zsh). Also, infers shell type automatically. Fixes #20 

Thanks @yardnsm for the `zsh` snippet!

**Note**: in order to use it, everyone who uses fnm needs to add double-quotes around the "env":
```diff
- eval `fnm env --multi`
+ eval "`fnm env --multi --use-on-cd`"
```

for Fish we moved to use `source` instead of `eval`:

```diff
- eval (fnm env --multi)
+ fnm env --multi --use-on-cd | source
```


executable/Env.re
```diff
@@ -25,8 +24,66 @@ let rec makeTemporarySymlink = () => {
...
+        fnm use
+      elif [[ -f .nvmrc && .nvmrc ]]; then
+        echo "fnm: Found .nvmrc"
+        fnm use
+      fi
+    }
+
+    alias cd=__fnmcd
+  |}
+  | Fish => {|
```
Schniz: Hey @elliottsj! do you want to take a look? I know it works - but do you think this is a good way to do it on fish?

elliottsj: Looks good to me! It's a little beyond my expertise, but from a cursory Google search, `--on-variable PWD` seems like a great way to do this.

PR #79, opened 2019-03-06
Guard from more non-existent directories errors

### LINKED ISSUE #75: `fnm use` fails if node version doesn't exist
Similar to #11, running `fnm use` when no node versions have been installed results in the same unfriendly error message and 125 error code:
```
$ fnm use
     Unix.Unix_error(Unix.ENOENT, "opendir", "/Users/<username>/.fnm/node-versions")
     Raised by primitive operation at file "src/unix/lwt_unix.cppo.ml", line 235, characters 14-31
     Re-raised at file "library/Fs.re", line 5, characters 2-441
     Re-raised at file "library/Versions.re", line 38, characters 4-262
     Re-raised at file "library/Versions.re", line 294, characters 2-497
     Re-raised at file "executable/Use.re", line 15, characters 2-1023
     Re-raised at file "src/core/lwt.ml", line 2987, characters 20-29
     Called from file "src/unix/lwt_main.ml", line 26, characters 8-18
     Called from file "cmdliner_term.ml", line 25, characters 19-24
     Called from file "cmdliner.ml", line 116, characters 32-39
$ echo $?
125
```

As a user I'd expect `fnm use` to install the version in case it didn't exist before. Please forgive if this is intended as my expectations come from using `nvm` before 😁; in such case, providing a friendlier message and pointing to `fnm install` would help.

Schniz:
Catch more Unix 125 errors, Fixes #75:

> I believe the error happens when you have zero node installations handled by fnm. Try removing those versions you have (or moving the `~/.fnm/node-versions/` folder somewhere else for that matter) and you should be able to reproduce the problem.

Thanks @julen for the report!

### DIFF
--- a/library/Fs.re
+++ b/library/Fs.re
@@ -35,4 +35,10 @@ let writeFile = (path, contents) => {
   Lwt.return();
 };
 
+let exists = path => {
+  try%lwt (Lwt_unix.file_exists(path)) {
+  | Unix.Unix_error(_, _, _) => Lwt.return(false)
+  };
+};
+
 let realpath = Filename.realpath;
--- a/library/Versions.re
+++ b/library/Versions.re
@@ -36,9 +36,13 @@ module Local = {
   let getLatestInstalledNameByPrefix = prefix => {
     open Lwt;
     let%lwt versions =
-      Fs.readdir(Directories.nodeVersions)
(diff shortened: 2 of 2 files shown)


julen:
Please note the error code is 125, not 127. Thanks for the quick fix!

Schniz:
whoops! sorry and thanks 😄

PR #88, opened 2019-03-26
Windows support

ulrikstrid:
This PR makes fnm build on Windows. It shouldn't change any behavior at this point and will not run on Windows at this point.


Schniz:
oh my god

👏 ❤️

ulrikstrid:
I think this PR should be ready for review/merge. I can rebase it to 1 commit if that is preferable.

As we talked about I added a `windows.json` that extends the `package.json` and then we can use windows with `esy --project windows ...`. The current version of pesy that is used can't run on Windows so I opted to run dune directly instead.

PR #91, opened 2019-04-10
Fix `fnm env` for fish shell.

hwartig:
After running `fnm env --multi | source` in fish I noticed that running `fnm use` still relinks `~/.fnm/current` rather then the folder in tmp that it is supposed to relink in the --multi case.

I realised that `fnm` cannot see the `FNM_...` variables as they are not exported (using -x) when they get set.

> ~/> set --show FNM_DIR
> $FNM_DIR: not set in local scope
> $FNM_DIR: set in global scope, unexported, with 1 elements
> $FNM_DIR[1]: length=17 value=|/Users/harry/.fnm|
> $FNM_DIR: not set in universal scope

So I added `-gx` to the `set` commands for fish so that the variables can be picked up by later runs of `fnm`. There are cases where the setup current code already works however it doesn't when the `FNM_...` variables haven't been exported globally before.

### DIFF
--- a/executable/Env.re
+++ b/executable/Env.re
@@ -104,12 +104,12 @@ let run = (~forceShell, ~multishell, ~nodeDistMirror, ~fnmDir, ~useOnCd) => {
     )
     |> Console.log;
   | Fish =>
-    Printf.sprintf("set PATH %s/bin $PATH;", path) |> Console.log;
-    Printf.sprintf("set %s %s;", Config.FNM_MULTISHELL_PATH.name, path)
+    Printf.sprintf("set -gx PATH %s/bin $PATH;", path) |> Console.log;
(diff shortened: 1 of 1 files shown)


Schniz:
Hey, and thanks! Too bad that my tests haven't caught that. If you think this is testable in some way, I'd love to get a PR that tests that too!

Thanks for the lovely fix. I'm gonna publish this soon 😄

hwartig:
Hey thanks for the quick merge. I didn't really try coming up with a good test as my capabilities in the used language are not existent. However I see this as a case for an integration test with the following steps.

1. Install `fnm` with no additional config passed to the installer (fish is used as the shell)
2. Reload fish shells config
3. run `fnm use v10.0.0` and assert `node -v` is correct
4. run `fnm use v11.0.0` and assert `node -v` is correct 

Step e is were I noticed things were wrong but depending on the actual versions (and the default version for node) it might already have failed at step 3.

PR #93, opened 2019-04-15
Add support for log level (closes #33)

### LINKED ISSUE #33: fnm use loudness
2 lines seems very verbose with message that looks like a debug info:

```
Linking /Users/Pavel/.fnm/current to /Users/Pavel/.fnm/node-versions/v10.15.1/installation
Using v10.15.1
``` 

Maybe something like:

`fnm using node v10.15.1` would be enough as default?

ohana54:
Supports "quiet", "error" and "all".

### DIFF
--- a/library/LogLevel.re
+++ b/library/LogLevel.re
@@ -1,0 +1,19 @@
+type t =
+  | Quiet
+  | Error
+  | All;
+
+let toString = logLevel =>
+  switch (logLevel) {
+  | Quiet => "quiet"
+  | Error => "error"
+  | All => "all"
+  };
+
+let fromString = logLevelString =>
+  switch (logLevelString) {
+  | "quiet" => Quiet
+  | "error" => Error
+  | "all" => All
+  | _ => failwith("Unsupported level: " ++ logLevelString)
+  };
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Lists the installed Node versions.
 
 Lists the Node versions available to download remotely.
 
-### `fnm env [--multi] [--shell=fish|bash|zsh] [--node-dist-mirror=URI] [--use-on-cd] [--base-dir=DIR]`
+### `fnm env [--multi] [--shell=fish|bash|zsh] [--node-dist-mirror=URI] [--use-on-cd] [--base-dir=DIR] [--log-level=quiet|error|all]`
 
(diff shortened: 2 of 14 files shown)


library/LogLevel.re
```diff
@@ -0,0 +1,19 @@
+type t =
```
Schniz: Consider moving this file to Logger

ohana54: That's where it was initially, but I had a cyclic dependency `Config -> Logger -> Config` because `Config` needs the log level and `Logger` needs the config value in runtime

ohana54:
I added the flag to the readme, I didn't elaborate as I think it's self explanatory, but let me know otherwise.

Schniz:
Sorry I haven’t looked at it yet. I’m on vacation 🇵🇱

library/LogLevel.re
```diff
@@ -0,0 +1,19 @@
...
+  | Error => "error"
+  | All => "all"
+  };
+
+let fromString = logLevelString =>
+  switch (logLevelString) {
+  | "quiet" => Quiet
+  | "error" => Error
+  | "all" => All
+  | _ => All
```
Schniz: I'm torn between `_ => All` to `_ => failwith("Unsupported level: " ++ logLevelString)`. What do you think? should we fail?

Schniz: p.s. just came back from vacation 😜

ohana54: I agree. I initially operated according to another tool that didn't fail on invalid level, but now that I think of it again we should fail.
I'll make the change, although it will take me a couple of days to get back to the code and modify+add test

Schniz: That's cool 😄
Thanks 😸

ohana54: As I understand the code, I think this code path is unreachable, since CmdLiner already does the validation according to this: https://github.com/ohana54/fnm/blob/497d8e1351cd0a8ad4e0c47baa95db2a934c8716/executable/FnmApp.re#L205-L207

I will still change it to fail, in case this function will be used in a different code path in the future, but I can't test it with an e2e test (I can write a unit for this function only if that's preferred)

Schniz: You’re right. Tests isn’t necessary in this case. Sorry!

ohana54: It's all good :)
Pushed the fix.

PR #99, opened 2019-05-17
fix EACCES error when installing an already downloaded version

tatchi:
This should fix an issue that arises we delete a version (by deleting the folder within the `node-versions` folder but keeping the downloaded `.tar.xz` file) and trying to reinstall it.

This was happening due to some access permission on the `.tar.xz` file. I tried to use other permissions when creating the file:
``` reason
Lwt_unix.openfile(path, [Unix.O_RDWR, Unix.O_CREAT], 777)
```

But didn't really understand it 😝 I finally ended up by using `Lwt_io` instead of `Lwt_unix` like [here](https://github.com/fastpack/fastpack/blob/master/FastpackUtil/FS.re#L81) and it seems to fix the issue 🎉

### DIFF
--- a/library/Fs.re
+++ b/library/Fs.re
@@ -26,12 +26,11 @@ let readdir = dir => {
 };
 
 let writeFile = (path, contents) => {
-  let%lwt x = Lwt_unix.openfile(path, [Unix.O_RDWR, Unix.O_CREAT], 777);
-  let%lwt _ =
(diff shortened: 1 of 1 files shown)


Schniz:
Very cool! thanks. I just did whatever worked - so I'm sorry there was a bug in the first place. Thanks for the cool fix 😄 🌮

PR #101, opened 2019-05-21
Move from base to core

ulrikstrid:
Vendored the Path library from [reason-native](https://github.com/facebookexperimental/reason-native/tree/master/src/path) for cross-platform realpath


library/Fs.re
```diff
@@ -40,7 +38,7 @@ let exists = path => {
   };
 };
 
-let realpath = Filename.realpath;
+let realpath = s => Path.absoluteExn(s) |> Path.toString;
```
ulrikstrid: Make sure this is the same thing, this will throw if it's not a path string

library/Versions.re
```diff
@@ -176,13 +176,21 @@ let getCurrentVersion = () => {
   | installationPath =>
     let fullPath = Filename.dirname(installationPath);
     Some(
-      Local.{fullPath, name: Core.Filename.basename(fullPath), aliases: []},
+      Local.{
```
ulrikstrid: And make sure this is the same thing as well

library/Versions.re
```diff
@@ -217,10 +225,10 @@ let getRemoteVersions = () => {
   let%lwt installedVersions = Remote.getInstalledVersionSet();
 
   versions
-  |> Core.List.filter(~f=x =>
+  |> Base.List.filter(~f=x =>
```
Schniz: I have zero idea why I have used `Core.List` for simply mapping and filtering, but what do you think if we're still here dropping this? `Base.List.map |> List.sort` feels weird

library/Versions.re
```diff
@@ -13,7 +13,7 @@ let compare = (v1, v2) =>
   switch (parseSemver(v1), parseSemver(v2)) {
   | (Some(v1), Some(v2)) => Semver.compare(v1, v2)
   | (None, _)
-  | (_, None) => - Core.String.compare(v1, v2)
+  | (_, None) => - Base.String.compare(v1, v2)
```
Schniz: Also, pretty sure `String.compare` works without `Base`

library/Fs.re
Schniz: This is not what `realpath` does, and I believe this is what causing the build errors.
`realpath` is like `Unix.readlink`, but it doesn't throw if it can't reach the destination: http://man7.org/linux/man-pages/man1/realpath.1.html

Schniz: Maybe I can make it better in another PR - assume `system` version if there's an error reading the link to the node version (because I use `/dev/null/installation` as the target 😜)

Schniz:
this is some amazing work! I hope it will get us closer to Windows support 🤩

Schniz:
Ohhhhhhh yeahhhhhhhhh

PR #102, opened 2019-05-22
Implement `realpath` instead of binding to C library

Schniz:
For future cross platform support
cc @ulrikstrid


library/Versions.re
```diff
@@ -172,24 +176,17 @@ let endsWith = (~suffix, str) => {
...
-  switch (Fs.realpath(Directories.currentVersion)) {
-  | installationPath =>
+  switch%lwt (Fs.realpath(Directories.currentVersion)) {
+  | Missing(x) when x == Directories.currentVersion => Lwt.return_none
+  | Missing(_) => Lwt.return_some(Local.systemVersion)
+  | Exists(installationPath) =>
     let fullPath = Filename.dirname(installationPath);
-    Some(
+    Lwt.return_some(
       Local.{fullPath, name: Core.Filename.basename(fullPath), aliases: []},
```
ulrikstrid: Still using `Core.Filename` here which is a problem on Windows

Schniz: Oh, I can take the changes you've made with @jordwalke's (and @reason-native's) `Path.re` and apply them here too 😄 But I don't want to make conflicts for #101

ulrikstrid: Either you could just incorporate what I did in #101 and close my PR or we can chat on Discord to figure this out 😄

PR #106, opened 2019-05-30
Add `default`, as a shortcut for `alias default`

### LINKED ISSUE #73: Feature requests / wishlist
thanks for the great version manager first of all!
just a list of features i'd like to have in an upcoming release 😄 

## features
- `fnm uninstall <version>`
- `fnm default <version>` to set a default version
- ~~auto select right node version if `.nvmrc` `.node-version` file exists~~ (sry it's already in progress https://github.com/Schniz/fnm/pull/68 👀 )

thanks again 😄

dangdennis:
Fixes #73 

Hey Schniz, this is the draft PR for the default command. Is this something you still want?

How do I actually run your feature tests? Below are my attempts:
 
`feature_tests/run.sh` => "No binary supplied!"
`feature_tests/aliases/run.sh` => "feature_tests/aliases/run.sh: Permission denied"

### DIFF
--- a/executable/FnmApp.re
+++ b/executable/FnmApp.re
@@ -3,6 +3,7 @@ let version = Fnm.Fnm__Package.version;
 module Commands = {
   let use = (version, quiet) => Lwt_main.run(Use.run(~version, ~quiet));
   let alias = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
+  let default = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
   let listRemote = () => Lwt_main.run(ListRemote.run());
   let listLocal = () => Lwt_main.run(ListLocal.run());
   let install = version => Lwt_main.run(Install.run(~version));
@@ -170,6 +171,33 @@ let alias = {
   );
 };
 
+let default = {
+  let doc = "Alias a version as default";
+  let man = [];
+  let sdocs = Manpage.s_common_options;
+
+  let selectedVersion = {
+    let doc = "The version to be aliased as default";
+    Arg.(
+      required
+      & pos(0, some(string), None)
+      & info([], ~docv="VERSION", ~doc)
+    );
+  };
+
+  (
+    Term.(const(Commands.alias) $ selectedVersion $ const("default")),
+    Term.info(
+      "default",
+      ~version,
+      ~doc,
+      ~exits=Term.default_exits,
+      ~man,
+      ~sdocs,
+    ),
+  );
+};
+
 let env = {
   let doc = "Show env configurations";
   let sdocs = Manpage.s_common_options;
@@ -268,6 +296,6 @@ let defaultCmd = {
 let _ =
   Term.eval_choice(
     defaultCmd,
-    [install, uninstall, use, alias, listLocal, listRemote, env],
+    [install, uninstall, use, alias, default, listLocal, listRemote, env],
   )
   |> Term.exit;
--- a/executable/FnmApp.re
+++ b/executable/FnmApp.re
@@ -3,7 +3,8 @@ let version = Fnm.Fnm__Package.version;
 module Commands = {
   let use = (version, quiet) => Lwt_main.run(Use.run(~version, ~quiet));
   let alias = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
-  let default = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
+  let default = version =>
+    Lwt_main.run(Alias.run(~name="default", ~version));
   let listRemote = () => Lwt_main.run(ListRemote.run());
   let listLocal = () => Lwt_main.run(ListLocal.run());
   let install = version => Lwt_main.run(Install.run(~version));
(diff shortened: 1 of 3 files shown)


executable/FnmApp.re
```diff
@@ -3,6 +3,7 @@ let version = Fnm.Fnm__Package.version;
 module Commands = {
   let use = (version, quiet) => Lwt_main.run(Use.run(~version, ~quiet));
   let alias = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
+  let default = (version, name) => Lwt_main.run(Alias.run(~name, ~version));
```
dangdennis: The shorter way would've been to set `~name="default"` but that doesn't look or feel right.

executable/FnmApp.re
```diff
@@ -170,6 +171,33 @@ let alias = {
   );
 };
 
+let default = {
+  let doc = "Alias a version as default";
+  let man = [];
+  let sdocs = Manpage.s_common_options;
+
+  let selectedVersion = {
+    let doc = "The version to be aliased as default";
```
dangdennis: I can't believe you started learning Reason and Ocaml late 2018 from your given articles. There's a lot to take in, especially just trying to understand a library like lwt and cmdliner, on top of setting up the build system.

Schniz:
> How do I actually run your feature tests?

This is totally my bad. The way to run them is by supplying the `fnm` binary, which is like:

```bash
esy build
./feature_tests/run.sh _build/default/executable/FnmApp.exe
```

executable/FnmApp.re
Schniz: I normally don't try to understand 100% of the things. This is why I use `pesy` here instead of pure Dune, but I also used Dune in another project where I tried to learn native OCaml/Reason.

Lwt feels a lot like JS promises, and I read a lot about functional programming and these monadic interfaces for a couple of years. This is a very big thing to learn but the `ppx_lwt` makes it easier, just like async await in JS

Cmdliner is very hard and I just read and copied from the docs until it worked! 😜  I read on the Reason Discord that there's a `ppx_deriving_cmdliner` which is interesting

executable/FnmApp.re
Schniz: I totally understand your take on this, but considering how hard is it for a new contributor (and new Reason user) to understand Cmdliner, I think it's better to set the `default` magic string here and not in the Cmdliner bloat. This module (`Commands`) is easy to read and pretty much sums everything up.

And it's not that we don't have the magic `default` string here, it is just written in the weird land of Cmdliner, where it's harder to follow on things 😜 

The other approach was to remove the `default` function completely because it is exactly like the `alias command`, and use the `const("default")` you have used into `Commands.alias`.

However, I think your choice of making a new command looks much nicer. Every command exposes a function in the `Commands` module. Cmdliner is just something we have to use in order to make good CLIs, but it's just an adapter 😄 

What do you think?

feature_tests/aliases/run.sh
```diff
@@ -7,8 +7,10 @@ fnm install 8.11.3
 
 fnm alias 8.11.3 oldie
 fnm alias 6.11.3 older
+fnm default 6.11.3 
 
 VERSIONS_INSTALLED=$(fnm ls)
 
 echo "$VERSIONS_INSTALLED" | grep 8.11.3 | grep oldie
 echo "$VERSIONS_INSTALLED" | grep 6.11.3 | grep older
+echo "$VERSIONS_INSTALLED" | grep 6.11.3 | grep default
```
Schniz: oh yes

Schniz:
Thanks for this PR! I'm sure people will be happy with it 😄 
Awesome work understanding how to work with Cmdliner, by the way! 🌮 🌵

executable/FnmApp.re
dangdennis: After much thinking and mulling in what you said, I agree that the magic string in the `Commands` module is very suitable. 

But having each `Commands` expose a new command with its own function is really intuitive. It also means we can customize CLI helper comments and whatnot for each command without having to overlap, say, `alias` and `default`. 

:D This is fun.

executable/FnmApp.re
dangdennis: Ah thanks for the quick summary. Reread the lwt docs and they clearly make sense now. 

And I finally understand what ppx are now too. They're like decorators in JS, or method attributes in C#. LOL.

dangdennis:
Darn, if we use the magic string default, the `Commands.default` breaks into a second line. Totally superfluous issue. Is this a prettier or eslint setting you set up somewhere?

Schniz:
> Is this a prettier or eslint setting you set up somewhere?

I use `refmt` which works like Prettier, or, actually, [the other way around](https://jlongster.com/A-Prettier-Formatter) 😉
https://github.com/Schniz/fnm/blob/cd0cb1aaa29a3f853f166d4159638c13ccfafff4/package.json#L122-L125

The linebreak doesn't bother me 😄

executable/FnmApp.re
Schniz: Most of them are like decorators or method attributes, but they work at compile time, unlike JS (not sure about C# though) - including type errors, type inference and autocomplete! ❤️

PR #118, opened 2019-07-01
Add note about upgrading 

pavelloz:


README.md
```diff
@@ -20,6 +20,7 @@
 
 ## Installation
 
+
```
pavelloz:
```suggestion
```

README.md
```diff
@@ -36,6 +36,10 @@ For `bash`, `zsh` and `fish` shells, there's an [automatic installation script](
 curl -fsSL https://github.com/Schniz/fnm/raw/master/.ci/install.sh | bash
 ```
 
+### Upgrade 
+
+You can upgrade `fnm` using the same command as you do to install it.
```
Schniz: What do you think about suggesting the `--skip-shell` option for upgrading?

pavelloz: Ouh, yeah, i had to deduplicate the `.profile` entry, so probably this is a must-have to be "oneliner" as I intended to be :)

Schniz: We can also look for `# fnm` in `$CONF_FILE` and just skip if we find it

https://github.com/Schniz/fnm/blob/671c833518c4ecfb7bbe28b5debcedf120332cc6/.ci/install.sh#L83

pavelloz: That would be a killer (especially for lazy ppl, which is most of us I suppose)

Having said that, while I remove comments because of `.rc` files get big quickly with all the comments, so i guess the combination of both (note about --skip-shell and sensible default about not duplicating it anyways) would be safest.

Schniz: So can you please add the `--skip-shell` part to the PR? 😄

PR #123, opened 2019-07-04
Fix typo in successfully

pavelloz:
SVG diff looks scary, but only because Github is pretty weak in diffing - only one letter has changed. I noticed you have automated generation of this SVG, but since its SVG, it should work just as good :)


Schniz:
haha! great catch!
no need to update the SVG because it is generated on every new version 😸

PR #146, opened 2019-10-02
Add support for `lts/*`

### LINKED ISSUE #137: Support `lts/*` in .nvmrc files
Currently, `lts/*` is considered an acceptable version by `nvm`. It knows to treat this as the latest LTS version. However, `fnm` doesn't seem to understand this syntax even though it's a valid .nvmrc config. Would it be possible to support this?

Schniz:
This should fix #137.

Apparently, [theres a `json` file in the Node dist](https://nodejs.org/dist/index.json) that holds all the versions. This can be queried against to find the latest LTS, but can also be used to get any version we want. It can be cached locally and it should work better than traversing the HTML nodes as we do right now.

This PR doesn't implement this but starting to use an uncached index.json. I believe that in a future PR I'll migrate everything to use that file!


feature_tests/run.sh
```diff
@@ -19,7 +19,7 @@ rm -rf $TEMP_DIR_BASE
 mkdir $TEMP_DIR_BASE $TEMP_BINARY_PATH
 cp $BINARY $TEMP_BINARY_PATH/fnm
 
-for test_file in $DIRECTORY/*/run.sh; do
+for test_file in $DIRECTORY/latest-lts/run.sh; do
```
thomsj: Was this a mistake, @Schniz?

Schniz: indeed! whoops!

PR #148, opened 2019-10-08
Make `env` smoke test pass when run with fish

thomsj:
Ensure esy `test` script runs `TestFnm.exe` as child of bash process, as
`env` smoke test fails when parent process is fish.

### DIFF
--- a/test/SmokeTest.re
+++ b/test/SmokeTest.re
@@ -11,7 +11,7 @@ describe("Smoke test", ({test, _}) => {
   });
 
   test("env", ({expect}) => {
-    let env = run([|"env"|]) |> redactSfwRoot;
+    let env = run([|"env", "--shell=bash"|]) |> redactSfwRoot;
     expect.string(env).toMatchSnapshot();
   });
 });


package.json
```diff
@@ -73,7 +73,7 @@
     "update-fnm-package": "node ./.ci/prepare-fnm-package.js",
     "verify-fnm-package": "node ./.ci/prepare-fnm-package.js --fail-on-difference",
     "bootstrap": ".ci/bootstrap",
-    "test": "esy x TestFnm.exe",
+    "test": "bash -c '(esy x TestFnm.exe)'",
```
Schniz: What do you think about changing the test to use `--shell=bash` explicitly instead of implying it here?

thomsj: 😂

That's what I did first, but I thought you would want to keep testing `System.Shell.infer()`. Is that tested elsewhere?

thomsj: I suppose it is tested in the feature tests since they don't include `--shell=bash` explicitly. Would you prefer `--shell=bash`?

Schniz: I think that yes. This is a smoke test, not a full-blown test. I'd rather unit-test it with a feature test, where lots of things are implicit

Schniz:
Looks great! Thanks!

PR #149, opened 2019-10-08
Update to v6.17.1 in `partial_semver` feature test

thomsj:
Update node 6 version to final/end-of-life version.


thomsj:
Do you think `exit 1` should be added?

Schniz:
@thomsj ah, yes!

PR #150, opened 2019-10-08
Add missing log level to `env` output for Fish shell

thomsj:
Refactor and add new type, used by records for POSIX and fish, to ensure
parity between shells.

### DIFF
--- a/executable/Env.re
+++ b/executable/Env.re
@@ -1,5 +1,55 @@
 open Fnm;
 
+module ExportShellVarCmdGetters = {
+  type exportCmdGetter = {
+    getPathCmd: string => string,
+    getMultiShellPathCmd: string => string,
+    getFnmDirCmd: string => string,
+    getNodeDistMirrorCmd: string => string,
+    getLogLevelCmd: string => string,
+  };
+
+  let posixExportCmdGetter: exportCmdGetter = {
+    getPathCmd: path => Printf.sprintf("export PATH=%s/bin:$PATH", path),
+
+    getMultiShellPathCmd: path =>
+      Printf.sprintf("export %s=%s", Config.FNM_MULTISHELL_PATH.name, path),
+
+    getFnmDirCmd: fnmDir =>
+      Printf.sprintf("export %s=%s", Config.FNM_DIR.name, fnmDir),
+
+    getNodeDistMirrorCmd: nodeDistMirror =>
+      Printf.sprintf(
+        "export %s=%s",
+        Config.FNM_NODE_DIST_MIRROR.name,
+        nodeDistMirror,
+      ),
+
+    getLogLevelCmd: logLevel =>
+      Printf.sprintf("export %s=%s", Config.FNM_LOGLEVEL.name, logLevel),
+  };
+
+  let fishExportCmdGetter: exportCmdGetter = {
+    getPathCmd: path => Printf.sprintf("set -gx PATH %s/bin $PATH;", path),
+
+    getMultiShellPathCmd: path =>
+      Printf.sprintf("set -gx %s %s;", Config.FNM_MULTISHELL_PATH.name, path),
+
+    getFnmDirCmd: fnmDir =>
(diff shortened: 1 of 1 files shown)


executable/Env.re
```diff
@@ -1,5 +1,55 @@
...
+        nodeDistMirror,
+      ),
+
+    getLogLevelCmd: logLevel =>
+      Printf.sprintf("set -gx %s %s", Config.FNM_LOGLEVEL.name, logLevel),
+  };
+};
+
 let symlinkExists = path =>
   try%lwt (Lwt_unix.lstat(path) |> Lwt.map(_ => true)) {
```
thomsj: I didn't notice that you had merged the reformatted code, before I committed and pushed, so I had purposefully left in the space after `try%lwt` here.

Schniz:
I understand what is the refactor, but I'm not sure what is the bug that it fixes.

I wonder what's the "ocaml"-ish way to do that. Would that be functors:

```reason
module EnvScript = (Shell: {
  let declarePathVariable: unit => string;
  // or even
  let declareEnvVar: unit => string;
}) => {
  let getShellScript = (~useOnCd) => {
    Shell.getStuff(...) ++ "\n" ...
  }
}
```

I once had an experiment with building lisp-ish declarations in Reason. I have migrated it to Sketch.sh: https://sketch.sh/s/tgificrbDSiA2KEcUqgxwv/

Anyway, this current PR makes lots of sense in terms of maintainability. I'd be happy to merge it — I just want to know which bug it fixes in Fish so we can document that! 😄

executable/Env.re
```diff
@@ -90,38 +140,19 @@ let run =
     multishell
       ? makeTemporarySymlink() : Lwt.return(Directories.globalCurrentVersion);
 
-  switch (shell) {
```
thomsj: When the log level was added to `Env.re` it was only added for bash and zsh, so with fnm 1.15.0 you can't change the log level with fish, since the log level isn't printed, and therefore isn't sourced.

executable/Env.re
```diff
@@ -90,38 +140,19 @@ let run =
...
-    |> Console.log;
-    Printf.sprintf("set -gx %s %s;", Config.FNM_DIR.name, fnmDir)
-    |> Console.log;
-    Printf.sprintf(
-      "set -gx %s %s",
-      Config.FNM_NODE_DIST_MIRROR.name,
-      nodeDistMirror,
-    )
-    |> Console.log;
-  };
```
thomsj: When the log level was added to `Env.re` it was only added for bash and zsh, so with fnm 1.15.0 you can't change the log level with fish, since the log level isn't printed, and therefore isn't sourced.

thomsj:
I still don't really know Reason (or OCaml), so I'm happy for you to change this if there is a better way.

Schniz:
> I still don't really know Reason (or OCaml)

Looks like you know Reason to me (and most of all, problem-fixing 😄)
Now that I understand the bug, seems that my first suggestion won't solve the problem!

btw, I made stickers for fnm — so if you want to get some let me know your address over twitter DMs or email and I'll try to make it happen 😄 your contributions are awesome and I really appreciate them.

Schniz:
I'm trying to rerun the Windows build because it should pass.

thomsj:
> Looks like you know Reason to me (and most of all, problem-fixing)

Thanks 😊. Unfortunately, I have probably forgotten most of what I learned by reading the Reason docs, since I didn't read/write any Reason code for a few weeks afterwards. 🤦‍♂️

> btw, I made stickers for fnm — so if you want to get some let me know your address over twitter DMs or email and I'll try to make it happen

Cool! You don't need to send me any. I don't put stickers on my laptop or anything else, I'm ~~a bit~~ really OCD about this 🤦‍♂️🤷‍♂️😂. I'd like to see them though, are they on Discord or Twitter or anywhere else?

> your contributions are awesome and I really appreciate them.

You are too kind. 🤭😂

PR #152, opened 2019-10-08
Change refmt action to use CircleCI artifact

Schniz:
This depends on https://github.com/facebook/reason/pull/2462 and should show a use case.
I think we can also create a simple GitHub action for this to hide the complexity!

It drops CI time from 3:40m to less than 20s!


Schniz:
Closing for inactivity. There is no faster way to do that right now.

PR #158, opened 2019-10-11
Uncapitalise "Node" in `--multi` description

thomsj:
Convert "Node" to "node" to improve consistency, as "node" is used     
throughout the manpages.

### DIFF
--- a/executable/FnmApp.re
+++ b/executable/FnmApp.re
@@ -283,7 +283,7 @@ let env = {
   };
 
   let isMultishell = {
-    let doc = "Allow different Node versions for each shell";
+    let doc = "Allow different node versions for each shell";
     Arg.(value & flag & info(["multi"], ~doc));
   };
 


thomsj:
"Node" is used almost exclusively in the README, so perhaps either the README or manpages should be changed to be more consistent.

thomsj:
I was all set to say: ["the OpenJS Foundation call it **'node'**", until I noticed the first sentence.](https://openjsf.org/blog/2019/10/03/openjs-foundation-welcomes-first-project-node-version-manager-nvm/) 😂

I found a couple of their repos which refer to it as **'Node'**:

- https://github.com/nodejs/abi-stable-node
- https://github.com/nodejs/build (in the description)

You could argue that since it's run from the command-line it is more common to write **'node'**.

Either way, I don't think it matters much. Even if you just kept it the way it is currently it would be fine.

PR #176, opened 2019-12-16
Specify the release to the install script

chrisdaley:
I am currently having some issues in an automated environment where I cannot use v1.17.0 due to `Illegal instruction` problems.  Fnm works great for me with version 1.16.0.  I thought it would be nice to be able to specify the release of fnm to install with the script.

For installing latest, nothing changes.
```
curl -fsSL https://raw.githubusercontent.com/chrisdaley/fnm/master/.ci/install.sh | bash
```

Providing the `-r` flag downloads the specified release.
```
curl -fsSL https://raw.githubusercontent.com/chrisdaley/fnm/master/.ci/install.sh | bash -s -- -r v1.16.0
```

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -3,6 +3,7 @@
 set -e
 
 INSTALL_DIR="$HOME/.fnm"
+RELEASE="latest"
 
 # Parse Flags
 parse_args() {
@@ -23,6 +24,11 @@ parse_args() {
       echo "\`--force-install\`: I hope you know what you're doing." >&2
       FORCE_INSTALL="true"
       shift
+      ;;
+    -r | --release)
+      RELEASE="$2"
+      shift # past release argument
+      shift # past release value
       ;;
     *)
       echo "Unrecognized argument $key"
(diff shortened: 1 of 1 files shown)


Schniz:
Hi Chris! thanks for the contribution and for acting to make the issue go away. However, I don't think this is what we want to have. I believe we should get this issue fixed instead of helping people installing an older version.

Can you please try installing the artifacts I shared in #177 and let me know if that works for you? My CPU has the `adx` instruction set so I can't know if it fixes the issue or not.

chrisdaley:
The linux artifact shared in #177 seems to work in my environment.

Whether you accept the change or not is up to you.  For me, it solves a couple issues:
- testing if there is a regression in the current version is made easier because I can just switch back to the previous release pretty easily
- allows me to be on a specific version of fnm across my environments until I need to update.

Schniz:
> * allows me to be on a specific version of fnm across my environments until I need to update.

Makes sense. It's unfortunate that there isn't a standard way to produce installation scripts/installers. More abilities to the installation script = more potential bugs, but this looks good and simple change.

Schniz:
> The linux artifact shared in #177 seems to work in my environment.

Also, I'm happy! I hope it will work for more people, so I can release a new version with this fix.

chrisdaley:
Thanks for the accept.  You are correct, I now technically have one more version of things to manage.

PR #191, opened 2020-02-19
uninstall allow prefix (fix #121)

### LINKED ISSUE #121: `uninstall` usability
I can `fnm install 12`, but i cannot `fnm uninstall 10` even if there is clearly only one fitting match.

Additionally, i have to type `vXX.XX.X`, numbers are not enough.

I think behavior could be more user friendly (which in this case also means consistent across commands).

```
marketplace-nearme-example|master⚡ ⇒ fnm ls
The following versions are installed:
* system
* v10.15.3
* v12.2.0
* v12.3.0
* v12.5.0
* v12.6.0 (default)
marketplace-nearme-example|master⚡ ⇒ fnm uninstall 10 
The version 10 is not installed.
marketplace-nearme-example|master⚡ ⇒ fnm uninstall 10.15.3
The version 10.15.3 is not installed.
marketplace-nearme-example|master⚡ ⇒ fnm uninstall v10.15.3
     Unix.Unix_error(Unix.ENOENT, "stat", "/Users/pavel/.fnm/node-versions/v10.15.3/installation/lib/node_modules/serverless/node_modules/.bin/flat")
```

PS. Ill report bug about error above in different issue - it looks like it didnt uninstall 10 because of that.

tatchi:
Trying to fix #121. I still need to write the tests but so you can already have a look and give your comments.

The logic goes like this:

* If no version is matching => we print out a message (like we used to)
* If only one version is matching (because the user provided a specific version `6.17.1` or because the prefix match only one installed version) => the version get uninstalled
* If multiple installed versions are matched: we print out the matching installed versions and ask the user to re-run the command with the correct version to uninstall.

Note: it's not possible to uninstall using `latest` or `latest-v9.x`, ... But that's already the case so nothing has changed here.

### DIFF
--- a/executable/Uninstall.re
+++ b/executable/Uninstall.re
@@ -2,14 +2,20 @@ open Fnm;
 open Lwt.Infix;
 
 let run = (~version) => {
-  let version = Versions.format(version);
+  let%lwt installedVersions = Versions.getInstalledVersions();
 
-  let%lwt installedVersion =
-    Versions.getInstalledVersions()
-    >|= List.find_opt(x => Versions.Local.(x.name == version));
+  let formattedVersionName = Versions.format(version);
+  let matchingLocalVersions =
+    installedVersions
+    |> Versions.(
+         List.filter(v =>
+           isVersionFitsPrefix(formattedVersionName, Local.(v.name))
+           || v.name == formattedVersionName
+         )
+       );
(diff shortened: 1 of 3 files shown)


Schniz:
Thanks @tatchi! 💰

PR #192, opened 2020-02-19
Bump ocaml to 4.08

tatchi:
I updated OCaml to 4.08. I had also to update two others dependencies (`base` and `ppx_let`).

First time I do that so let's see if the CI pass 😆


tatchi:
It failed on Linux:

```
docker: Error response from daemon: pull access denied for schlez/fnm-static-binary, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```

@Schniz Do you have any idea ?

Schniz:
> @Schniz Do you have any idea ?

The docker image creation has failed. The `.ci/create-static-binary.sh` script that I wrote does not have `set -e` on the top, making failed commands just continue instead of failing the whole script

tatchi:
Youhhhouu 🎉🤗

Schniz:
Thanks @tatchi! 🤑 🌮

PR #193, opened 2020-02-20
Fix illegal instruction coming from `nocrypto`

Schniz:
`ocaml-nocrypto` uses hardware acceleration, this PR avoids it on CI.
For more details: https://github.com/mirleft/ocaml-nocrypto#illegal-instructions

Thanks for @ulrikstrid who always finds a way to fix fnm issues. ❤

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ RUN npm i -g esy@latest
 WORKDIR /app
 ADD . /app
 
+# https://github.com/mirleft/ocaml-nocrypto#illegal-instructions
+ENV NOCRYPTO_ACCELERATE false
+
 RUN jq '. | .buildDirs.executable.flags |= . + ["-ccopt", "-static"]' package.json > package.json.new && mv package.json.new package.json
 RUN npx esy i
 RUN npx esy verify-fnm-package
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,6 @@ RUN npm i -g esy@latest
 WORKDIR /app
 ADD . /app
 
-# https://github.com/mirleft/ocaml-nocrypto#illegal-instructions
-ENV NOCRYPTO_ACCELERATE false
-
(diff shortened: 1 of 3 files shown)


Schniz:
@Grohden would you mind test the following artifacts? (whatever fits your OS)

[fnm-macos - nocrypto.zip](https://github.com/Schniz/fnm/files/4231479/fnm-macos.-.nocrypto.zip)
[fnm-linux - nocrypto.zip](https://github.com/Schniz/fnm/files/4231480/fnm-linux.-.nocrypto.zip)

Grohden:
2020-02-20-134246_672x521_scrot
@Schniz Same thing I think (using linux btw)

ulrikstrid:
This fix could be removed by esy in the sandbox. There is a global environment option in esy that you can use for this case.

Schniz:
Thanks @ulrikstrid. @Grohden can you please try the following artifacts:

[fnm-linux - nocrypto.zip](https://github.com/Schniz/fnm/files/4233042/fnm-linux.-.nocrypto.zip)
[fnm-macos - nocrypto.zip](https://github.com/Schniz/fnm/files/4233045/fnm-macos.-.nocrypto.zip)

Grohden:
@Schniz same thing with those

Schniz:
> @Schniz same thing with those

Does fnm works for you if you build it locally?

It’s hard fixing this problem when I don’t have access to a Maine that fails with this. So I’m sorry for the time you can’t use fnm

Grohden:
@Schniz oh no problem - I think that it's a very specific problem that is hard to test fixes for. 

I recently got a new machine so now I'm able to use fnm without problems.. but I will not be able to build fnm for some time as I don't have access to the old machine due to our quarantine here...

Schniz:
> I recently got a new machine so now I'm able to use fnm without problems.. but I will not be able to build fnm for some time as I don't have access to the old machine due to our quarantine here... 

No problems. We can waited for the next person who has this problem. I’m not sure on what machines these instructions don’t work but we don’t need to fix things that no one needs. 

Hope you’re feeling well and everything is good! Stay safe

PR #201, opened 2020-02-26
Create alias from version range (fix #185)

### LINKED ISSUE #185: Cannot create an alias to a version range
With `nvm` I could do this:
```
nvm alias default 12
```
and have the default always point to the latest installed `12.x.y` version. With `fnm` this doesn't work, you have to provide a full version:
```
$ fnm default 12
Can't find a version installed in ~/.fnm/node-versions/v12/installation
$ ls ~/.fnm/node-versions/
v10.18.1 v12.14.1
```

In my experience, aliasing the default to a concrete Node version is not convenient. I want a specific major as different majors may have a breaking change but other than that, I want the latest.

tatchi:
Based on #194 to benefit from the `LocalVersionResolver` module. So I guess this PR should be merged first. This should fix #185 

To be consistent with the `use` command, we could propose to download and alias the version if not yet installed. What do you think?

### DIFF
--- a/executable/Alias.re
+++ b/executable/Alias.re
@@ -2,33 +2,32 @@ open Fnm;
 
 let run = (~name, ~version) => {
   let version = Versions.format(version);
-  let versionPath =
-    Filename.concat(
-      Filename.concat(Directories.nodeVersions, version),
-      "installation",
-    );
-  let%lwt versionInstalled = Lwt_unix.file_exists(versionPath);
+  let%lwt matchingLocalVersions =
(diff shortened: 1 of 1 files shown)


tatchi:
Fixed a regression bug caught by the tests 😄

mgol:
Any chance for a nearby release with this feature? :)

Schniz:
Ha! Sure. I’ll do it when I get to my computer so it’s either tomorrow or two days from now. Sorry!

PR #220, opened 2020-05-11
Add `current` command to retrieve the current Node version

vladimyr:
Display currently activated Node version.

### DIFF
--- a/executable/Current.re
+++ b/executable/Current.re
@@ -1,0 +1,11 @@
+open Fnm;
+
+let run = () => {
+  let%lwt currentVersion = Versions.getCurrentVersion();
+  switch (currentVersion) {
+  | Some({name}) => Console.log(name)
+  | _ => Console.log("none")
+  };
+
+  Lwt.return_ok();
+};
--- a/feature_tests/current/run.sh
+++ b/feature_tests/current/run.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+PATH="$(pwd)":$PATH # simulating a custom `node`
+
+eval "$(fnm env)"
+
+if [ "$(fnm current)" != "none" ]; then
+  echo "Expected currently activated version is not none!"
+  exit 1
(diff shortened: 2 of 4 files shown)


Schniz:
Cool! can we also add a small feature test? 😄

vladimyr:
> Cool! can we also add a small feature test? 😄

Shell or reason?

Schniz:
I like using shell because it shows how the user will interact with your feature. But you can use Reason too, as long as we're sure that I won't break your feature in the future 😄

vladimyr:
Feature tests added 🎉

feature_tests/current/run.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+PATH="$(pwd)":$PATH # simulating a custom `node`
```
Schniz, line 5: I'm not sure how it simulates a custom node

executable/Current.re
```diff
@@ -0,0 +1,11 @@
+open Fnm;
+
+let run = () => {
+  let%lwt currentVersion = Versions.getCurrentVersion();
+  switch (currentVersion) {
+  | Some({name}) => Console.log(name)
+  | _ => Console.log("none")
+  };
```
Schniz, line 8: So slick

feature_tests/current/run.sh
Schniz: but I'm merging this anyway, we can remove it later

vladimyr: > I'm not sure how it simulates a custom node

You are absolutely right. Simulating custom `node` here isn't needed and this doesn't do that anyway 🤦

executable/Current.re
vladimyr: I hope so, this is the first thing I'm doing in OCaml/Reason, ignoring obligatory Hello World, and definitely isn't much more advanced 😁

PR #225, opened 2020-06-03
Remove unused condition

joliss:
This just checks if ".node-version" is a non-empty string, which is
always true.

------

Was this meant to check if the file is non-empty? If so, I can send another pull request that `stat`'s the file.

### DIFF
--- a/executable/Env.re
+++ b/executable/Env.re
@@ -80,10 +80,10 @@ let printUseOnCd = (~shell) =>
     __fnmcd () {
       cd "$@"
 
-      if [[ -f .node-version && .node-version ]]; then
+      if [[ -f .node-version && -r .node-version ]]; then
         echo "fnm: Found .node-version"
         fnm use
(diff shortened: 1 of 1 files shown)


vladimyr:
@joliss Actually that is probably a typo. Compare it to `zsh` version that has following check:
```zsh
if [[ -f .node-version && -r .node-version ]]; then
  echo "fnm: Found .node-version"
  fnm use
elif [[ -f .nvmrc && -r .nvmrc ]]; then
  echo "fnm: Found .nvmrc"
  fnm use
fi
```
Seems like `-r` operator got accidentally omitted.

Schniz:
> Seems like `-r` operator got accidentally omitted.

yup!

Sorry @joliss that it took me a lot of time to comment back, for some reason I didn't see the email!

joliss:
Ah, thanks! I pushed this branch again with the missing `-r` added!

Schniz:
Thanks @joliss!

PR #227, opened 2020-06-12
Feature/support ash

mrtrom:
I wanted to install `fnm` on an alpine docker image, but got an error because `ash` is not supported by the installation script. This simple PR aims to solve that.

I also had another issue regarding the use of `$SHELL`:
```sh
/ # echo $SHELL


/ # basename $SHELL
basename: missing operand
Try 'basename --help' for more information.
```
$SHELL was empty when accessing via docker, so `basename` does not work providing an empty value. I solved this by doing `SHELL=($which ash)` on the image, I don't think that's an optimal solution but a workaround. 
Any thoughts about this? maybe providing a default install method (bash?) or printing a warning when no $SHELL is set 🤔

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -110,7 +110,7 @@ check_dependencies() {
 }
 
 setup_shell() {
-  CURRENT_SHELL=$(basename $SHELL)
+  CURRENT_SHELL=$(basename $0)
 
   if [ "$CURRENT_SHELL" == "zsh" ]; then
     CONF_FILE=$HOME/.zshrc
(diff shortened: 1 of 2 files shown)


Schniz:
Hey! sorry for not commenting on this!
thanks for the contribution.

Have you checked that fnm really works on `ash`? I wonder how the scripts would need to be changed. Can you take a look and add a simple test (like `fish` has)?

Here's the inference code:
https://github.com/Schniz/fnm/blob/bf5e3a70f333cdd228387a6c65abaf575bfe05ca/library/System.re#L30-L44

We might want to test for `-ash` and `ash` as well. Maybe to fall back to Bash syntax? would it work?

---

In other words, if we want to allow `ash` as an installation target, we need to verify that it works and we won't break it by adding another feature. Can you please assist with that? 🙏

Schniz:
> printing a warning when no $SHELL is set 🤔

That's a great idea too!

Schniz:
Hey there, since fnm was rewritten and it's now easier to make sure all the e2e tests work on other shells, we can try doing that. If `ash` syntax is the same as Bash, we can reuse most of the code.

Closing that for now, since I don't think adding an official supported shell without tests is a good idea. Please reopen if you think otherwise — discussion is always an option 😃

PR #244, opened 2020-10-01
Allow using homebrew to install with the installation script

Schniz:
I don't really know how to make Homebrew patch the shell profiles
so making the install script use Homebrew might work well.

This will make Homebrew the default installation mechanism for fnm, which is already the case, but now will be supported by the installation script.

This can be overridden and work the same way as the Linux installation the following way:

```bash
curl SCRIPT | bash -s -- --force-no-brew # or --force-install
```

This makes a great follow up question:
Is there a way to automate apt/yum/etc so we'll use real package managers instead of implementing our own for Linux boxes?

### DIFF
--- a/.ci/get_shell_profile.sh
+++ b/.ci/get_shell_profile.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+OS="$(uname -s)"
+
+case $1 in
+  "fish")
+    CONFIG_DIR="$(fish -c 'echo -n $__fish_config_dir')"
+    echo "${CONFIG_DIR-"$HOME/.config/fish"}/fish.config"
+    ;;
+  "zsh")
+    echo "$HOME/.zshrc"
+    ;;
+  "bash")
+    if [ "$OS" = "Darwin" ]; then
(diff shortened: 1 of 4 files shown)


waldyrious:
@Schniz Homebrew does work on Linux, as long as the formula code compiles the program from source, rather than downloading a precompiled binary. Couldn't the [fnm formula](https://github.com/Schniz/homebrew-tap/blob/master/Formula/fnm.rb) be adapted to that end?

Schniz:
> Couldn't the [fnm formula](https://github.com/Schniz/homebrew-tap/blob/master/Formula/fnm.rb) be adapted to that end?

I'll adapt it to favor downloading a prebuilt binary but compile when necessary. Now that the repo is Rust and not native Reason, it should be easier since it is more supported. Thanks!

waldyrious:
AFAIK a precompiled binary (bottle) is provided automatically by Homebrew. It's possible that that's only available for formulas in the [core repo](https://github.com/Homebrew/homebrew-core), but I see no reason not to migrate the formula there once it compiles fnm from source :)

Issue #222, opened 2020-05-22
Add a prebuilt ARM binary


mbeaudru:
Steps followed:

- Install fnm through the script: https://github.com/Schniz/fnm#using-a-script
- Restart the terminal

Issue:

- `~/.zshrc:110: exec format error: fnm` prompts after opening a new terminal
- Running `fnm` prompts `sh: exec format error: fnm`

Environment:

Raspberry Pi under raspbian

alx365:
Same here with archlinux arm on a Raspberry Pi. Please remove a arm version.

Schniz:
Sorry that it doesn't work, fellas!

I personally don't own a Raspberry Pi. Would you mind explaining to me what's wrong? Is it because it wasn't build for an ARM processor? What are the steps necessary to do that? Running on a different Docker image?

alx365:
So yes, the Pi uses the ARM architecture, What you will have to do, is either cross compile with whatever tool you are using, or just compile it on a arm device, such as your phone (probably a rather stupid idea), or running a arm OS in a virtual machine, that is fully emulated. 
By now i already looked in your README, and saw that you use "esy" to compile your program
I couldn't find anything online on how to cross compile with that particular compiler, in the 5 minutes i searched.

Schniz:
It looks like Dune can cross compile, not sure about ARM though. If someone wants to jump to the native development Discord and do that research, it could be awesome!

Schniz:
@mbeaudru do you mind if I change the title to "Allow to run on ARM devices"?

mbeaudru:
I just did it ! Thank you for working on this project by the way 👍

jaythomas:
9ba282a705c7a8f3d2dfbfb08d4c62720f600676 we got Rust now :eyes: ?? Is there anything I can do to help?  Not sure what's left but I have a Pinebook Pro running Arch Arm. I see `cargo build` runs without a hitch. Perhaps we can get a published crate too?

Schniz:
Looks like we can use https://github.com/rust-embedded/cross and create binaries for ARM and probably test everything with it as well. Maybe I have a Raspberry Pi 2 somewhere to test it with but I'm not sure where it is 😜  Would you want to collaborate on this?

> > I see `cargo build` runs without a hitch.

The fact it builds is awesome! but I think we still need to make a code change. Don't really know ARM terminology but I think we need to change this:

[...]

Issue #239, opened 2020-08-29
Arch detection override


takase1121:
I am running fnm on Alpine, meaning no glibc thus official node.js build can't be used. I set the appropriate mirror env variable to the unofficial one, but it still fails to install because my arch is x64 but the unofficial build has `-musl` appended onto that. Hopefully a flag to override that can be added

Schniz:
It is not currently supported. Sorry! But it is interesting.

Schniz:
> I set the appropriate mirror env variable to the unofficial one
would you mind sharing the unofficial mirror? 😄

takase1121:
Website: https://unofficial-builds.nodejs.org

Mirror: https://unofficial-builds.nodejs.org/download/

Schniz:
Potentially refactor for #239, most of this will be used in `cfg!` if statements instead of `#[cfg]` and maybe we'll have a PLATFORM constant that we can unit test, with the following possible values:

```
enum OsPlatform {
  Arm32,
  Arm64,
  X86,
  X64,
}
```

[Here are some docs regarding `#[cfg]` and `cfg!`](https://doc.rust-lang.org/stable/rust-by-example/attribute/cfg.html#cfg), and here's an example:

```rust
#[cfg(unix)]
fn this_fn_will_only_be_compiled_on_unix() {
  println!("Hello, Unix!");
}

#[cfg(not(unix))]
fn this_fn_will_not_be_compiled_on_unix() {
  println!("Hello, Windows!");
}

# Can be written like:

fn this_fn_will_always_compile() {
  if cfg!(unix) {
    println!("Hello, Unix!");
  } else {
    println!("Hello, Windows!");
  }
}
```

PR #274, opened 2020-10-17
Add ARM support

jaythomas:
This is the code from #222. I tested 64-bit arm (`aarch64`) and x64 but have no 32-bit devices. I'm fairly certain they work too though given the way the logic follows.

### DIFF
--- a/src/system_info.rs
+++ b/src/system_info.rs
@@ -8,12 +8,22 @@ pub fn platform_name() -> &'static str {
     "linux"
 }
 
-#[cfg(target_pointer_width = "32")]
+#[cfg(all(target_pointer_width = "32", target_arch = "arm"))]
+pub fn platform_arch() -> &'static str {
+    "armv7l"
+}
+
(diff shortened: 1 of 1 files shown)


Schniz (approved):
Thanks for doing that and ensuring that works. 👍 
I'm really excited to support ARM.

---

In order to really add that ARM is supported to the README, we'll gonna have to do the following:

* Add a CI step that will test with `cargo-cross`, and potentially add a `fnm-arm64.zip` and `fnm-arm32.zip` releases.
* Potentially refactor for #239, most of this will be used in `cfg!` if statements instead of `#[cfg]` and maybe we'll have a `PLATFORM` constant that we can unit test, with the following possible values:
  ```rust
  enum OsPlatform {
    Arm32,
    Arm64,
    X86,
    X64,
  }
  ```

---

(sorry if my comments are really long, that's a nice way to remember what I want to do, or maybe even get others to contribute. well, it's hacktoberfest, maybe I can ask for someone who's new to Rust to do the enum. I'm going to do that.)

PR #275, opened 2020-10-18
Fix moving node install out of tmp directory

jaythomas:
Addresses issue in #222 where `fnm install` doesn't work.

Per my comments on there, [`fs::rename`](https://doc.rust-lang.org/std/fs/fn.rename.html) cannot be used to move a directory across mount points, so the current code doesn't work on many Linux systems where `/tmp` is usually on a separate partition than `/home`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -348,6 +348,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "dircpy"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4479788c6c76674c1551ef44c953101554e7edadb5ea0920797a2280d87eb3d"
+dependencies = [
+ "log",
+ "walkdir",
+]
+
+[[package]]
 name = "dirs"
 version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -477,6 +487,7 @@ dependencies = [
  "clap",
  "colored",
  "csv",
+ "dircpy",
  "dirs",
  "duct",
  "embed-resource",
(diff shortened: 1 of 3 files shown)


Schniz (approved):
Woah, awesome! Thanks for fixing the bug. Didn't really saw the caveat of `fs::rename`. 👏  

For future reference, we can experiment with creating the directory in the same installation dir. This can be done with [TempDir::new_in](https://docs.rs/tempfile/3.1.0/tempfile/struct.TempDir.html#method.new_in), since copying can take a while. Using `new_in` will have some edge cases:
* the target can't be ~/.fnm/temp` because people can mount `~/.fnm/node-version` in a different file system. So the temp dir should be something like `~/.fnm/node-versions/.temp/` and I hope that'll work.
* It also means we need to clear or at least have a `fnm cache clear` command — because although the `TempDir` struct will remove all unused files, if the system was rebooted before doing that, we'll have dangling data.

Schniz:
btw, `DirectoryPortal` was one of the things I really liked I could do with Rust 😝  so thanks for making it better 😄

jaythomas:
If you want to change the existing `temp_dir` reference to point to `~/.fnm/node-versions/.temp/` that would certainly be more efficient than copying. Performance and simplicity are fnm's virtues afterall. Rather than creating a new `fnm cache clear` command and put maintenance on the user I think running a cleanup routine during the *next* `fnm install` would keep things simpler. That's my 2¢ (or 2 shekels :slightly_smiling_face:).

Just let me know what I can do to help, thanks!

Schniz:
> I think running a cleanup routine during the next fnm install would keep things simpler.

That's a good idea 😄 
If you're willing to try and create the temp directory inside `~/.fnm/node-versions` it would be awesome 😄 and it will make the code easier to understand too

Schniz:
Let's merge this and do the TempDir::new_in in a different PR.

PR #287, opened 2020-10-26
Remove `--multi` from install script

### LINKED ISSUE #286: fix: --multi is deprecated, but install script still includes it
Similar to #280. Installing with the provided script still includes it in the environment, which outputs the warning in the terminal for unaware users. From my short experience before addressing it, the warning appeared intermittently in the fish shell. Resolution for this would be to remove the `--multi` flag from the install script.

This should be the last place `--multi` is used in the source code, other than the deprecation warning. :blush: :+1:

jameschensmith:
Fixes #286.


Schniz:
Thanks!

PR #289, opened 2020-10-26
Remove logs from automatic version switching in `--use-on-cd`

### LINKED ISSUE #166: Could --log-level=quiet be quieter?
Hi there.

I'm using fnm with zsh and I'm very happy with it. Very fast and good.

I have the following in my `~/.zshrc`:

```shell
eval "$(fnm env --multi --use-on-cd --log-level=quiet)"
```

When I `cd` into a directory which has a `.node-version` file, fnm will notice and switch node version. At this point, it prints out:

```
fnm: Found .node-version
```

I would love if it didn't print out anything. That's my read on what "quiet" means. Apologies if I'm missing something.

maxjacobson:
Closes #166

Was glad to see this rewritten in rust. I've tested this locally in zsh, but I haven't run the tests. V open to feedback...

### DIFF
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -1,5 +1,4 @@
 use super::shell::Shell;
-use crate::log_level::LogLevel;
 use indoc::indoc;
 use std::path::PathBuf;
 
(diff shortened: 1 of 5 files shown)


src/shell/bash.rs
```diff
@@ -18,24 +19,37 @@ impl Shell for Bash {
...
-                    fi
-                }
-
-                alias cd=__fnmcd
-            "#
+    fn use_on_cd(&self, config: &crate::config::FnmConfig) -> String {
+        format!(
+            indoc!(
+                r#"
+                    __fnmcd () {{
```
maxjacobson, line 26: Needed to escape the curly braces, now that this is using `format!`

src/shell/shell.rs
```diff
@@ -6,6 +6,9 @@ pub trait Shell: Debug {
     fn set_env_var(&self, name: &str, value: &str) -> String;
     fn use_on_cd(&self, config: &crate::config::FnmConfig) -> String;
     fn into_structopt_shell(&self) -> structopt::clap::Shell;
+    fn echo_found(&self, _file_name: &str, _config: &crate::config::FnmConfig) -> Option<String> {
```
maxjacobson, line 9: perhaps this doesn't belong here at all, since some of the shells don't have logging at all...

Schniz:
Maybe we should just don't log in the shell? If you run with the `RUST_LOG=fnm=info` env variable, you'll get logs that say which file was found anyway..

Schniz:
Would you mind changing the PR to simply remove the "echo" statements? 😄

Schniz:
Oh wow, I just realized that you're the one opened #166! amazing!

maxjacobson:
> Would you mind changing the PR to simply remove the "echo" statements? 😄

much simpler! I like it.

> Oh wow, I just realized that you're the one opened #166! amazing!

🙂 yeah, just noticed it was rewritten in rust when I `brew upgrade`'d and my shell started printing out warnings about `--multi` being deprecated. Also, I needed to move the `--log-level` flag from:

```
eval "$(fnm env --multi --use-on-cd --log-level=quiet)"
```

To

```
eval "$(fnm  --log-level quiet env --use-on-cd)"
```

So I pulled up the repo to see what changed. 

Not a big deal, and it led me to realize it was rewritten in a language I understand enough to contribute to. Although, with this simpler solution, I probably could have done it in reasonml as well 🙈

Schniz:
Thanks for this!

> Although, with this simpler solution, I probably could have done it in reasonml as well 🙈

I could do it too but never got to do it either 🙈

PR #293, opened 2020-10-26
Add detailed error instead of panic regarding `fnm env`

Schniz:
This will provide more context for people who install fnm using Homebrew, because adding the installation docs as caveats is a misuse (https://github.com/Homebrew/homebrew-core/pull/63459)

also, panics are bad so we should avoid them and have some nice looking errors.

### DIFF
--- a/src/commands/use.rs
+++ b/src/commands/use.rs
@@ -24,6 +24,8 @@ impl Command for Use {
     type Error = Error;
 
     fn apply(self, config: &FnmConfig) -> Result<(), Self::Error> {
+        let multishell_path = config.multishell_path().context(FnmEnvWasNotSourced)?;
+
         let all_versions =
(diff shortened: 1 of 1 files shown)


src/commands/use.rs
```diff
@@ -134,4 +132,11 @@ pub enum Error {
         "Can't find version in dotfiles. Please provide a version manually to the command."
     ))]
     CantInferVersion,
+    #[snafu(display(
+        "{}\n{}\n{}",
+        "We can't find the necessary environment variables to replace the Node version.",
+        "  Have you set up your shell profile to evaluate `fnm env`?",
+        "Check out our documentation for more information: https://fnm.vercel.app"
```
Schniz, line 139: I really wish I could have used `indoc` or something here. This is ugly!

PR #307, opened 2020-10-28
Refer to homebrew/core formula instead of custom tap

jameschensmith:
Updated locations which reference the deprecated tap. Homebrew instructions will now relate to the `homebrew/core` formula.

I would have liked to add an automated Homebrew release strategy in this PR as well, but it seems that the current release strategy is done manually (i.e. outside of GitHub Actions (GHA) at least; please correct me if I'm wrong). A new issue outlining an automated strategy for pushes to the main branch done via GHA would definitely help get the ball rolling with this 😉 

Relates to Schniz/homebrew-tap#12.

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -70,7 +70,7 @@ set_filename() {
 
 download_fnm() {
   if [ "$USE_HOMEBREW" == "true" ]; then
-    brew install Schniz/tap/fnm > /dev/null
+    brew install fnm > /dev/null
   else
     if [ "$RELEASE" == "latest" ]; then
       URL="https://github.com/Schniz/fnm/releases/latest/download/$FILENAME.zip"
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -70,7 +70,7 @@ set_filename() {
 
 download_fnm() {
   if [ "$USE_HOMEBREW" == "true" ]; then
-    brew install fnm > /dev/null
+    brew install fnm
   else
     if [ "$RELEASE" == "latest" ]; then
       URL="https://github.com/Schniz/fnm/releases/latest/download/$FILENAME.zip"
--- a/.github/workflows/installation_script.yml
+++ b/.github/workflows/installation_script.yml
@@ -66,8 +66,8 @@ jobs:
       - run: "sudo apt-get install -y ${{ matrix.shell }}"
         name: Install ${{matrix.shell}} using apt-get
         if: matrix.setup.os == 'ubuntu'
-      - run: "brew install ${{ matrix.shell }}"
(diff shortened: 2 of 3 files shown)


jameschensmith:
Not sure why the test failed to find the `homebrew/core` formula 🤔  I can validate it's there:

```
$ brew info fnm
fnm: stable 1.22.2 (bottled), HEAD
Fast and simple Node.js version manager
https://fnm.vercel.app
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/fnm.rb
License: GPL-3.0-only
==> Dependencies
Build: rust ✘
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 13 (30 days), 13 (90 days), 37 (365 days)
install-on-request: 13 (30 days), 13 (90 days), 37 (365 days)
build-error: 0 (30 days)
```

.ci/install.sh
```diff
@@ -70,7 +70,7 @@ set_filename() {
 
 download_fnm() {
   if [ "$USE_HOMEBREW" == "true" ]; then
-    brew install Schniz/tap/fnm > /dev/null
+    brew install fnm > /dev/null
```
Schniz, line 73: Let's remove the `> /dev/null` because there are no caveats now 😄 
Then we'll see what Homebrew is telling us

jameschensmith:
@Schniz, this time, it worked for [`fish, macos`](https://github.com/Schniz/fnm/pull/307/checks?check_run_id=1320886182) but not [`bash, macos`](https://github.com/Schniz/fnm/pull/307/checks?check_run_id=1320886347). Only `fish, macos` was run the first time, because when it failed the other ones were skipped. There's got to be some caching happening in the background, which is making Homebrew fail the first time 🤔

Schniz:
I wonder if we should add a step (only for MacOS) that does `brew update` before installing. We already do `brew install SHELL`, we can turn it into a multiline expression and stick a `brew update` before that

jameschensmith:
> I wonder if we should add a step (only for MacOS) that does `brew update` before installing. We already do `brew install SHELL`, we can turn it into a multiline expression and stick a `brew update` before that

That sounds like a potential solution. I like it 🙂  I could change [these lines](https://github.com/Schniz/fnm/blob/v1.22.4/.github/workflows/installation_script.yml#L69-L71) to the following:

```
      - run: "brew update && brew install ${{ matrix.shell }}"
        name: Update then install ${{matrix.shell}} using Homebrew
        if: matrix.setup.os == 'macos'
```

jameschensmith:
@Schniz, it worked 😋 👍

Schniz (approved):
Amazing!

PR #315, opened 2020-10-31
Add `list` alias for `ls`

probablykasper:
In my opinion it would be sensible to add `fnm list` as an alias of `fnm ls`. At least for me, it's the natural command I think of for listing a version manager's installed versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,11 +152,11 @@ Activates `[VERSION]` as the current Node version. If no version provided, it wi
 
 Display currently activated Node version.
 
-### `fnm ls`
+### `fnm list`
 
 Lists the installed Node versions.
 
-### `fnm ls-remote`
+### `fnm list-remote`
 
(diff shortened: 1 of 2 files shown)


Schniz (approved):
Thanks. I hope it will be easier to understand.
Maybe this should be the name since it is more descriptive?

Would you also want to add `list-remote` as a human readable version of `ls-remote`? 😄

probablykasper:
@Schniz I think that would make sense, yeah. I know plenty of CLI tools using `list`, but never encountered `ls`.

Will change the names to `list` and `list-remote`, and keep `ls` and `ls-remote` as aliases :)

Schniz:
Sounds good. Thanks for the contribution 😁

probablykasper:
@Schniz Mo problem! Thanks for making fnm, it's pretty great

Issue #316, opened 2020-11-02
The --using-file command line argument is no longer present


alexjeffburke:
Hi there,

At my $work there are scripts calling `fnm` with a `--using-file` argument and it seems that as of the rust rewrite that is no longer supported. On that basis, there might be some compatibility breakage in the new version.

I'm not totally sure why we were using this argument when choosing the node version because it would seem to be default behaviour - my _guess_ is that the script calling `fnm` this wanted to make sure it bailed out if it was not able to use the node version specified in (in our case) `.nvmrc` rather than continue with some other default.

Thanks!

bjornua:
It seems like this flag is now the default behaviour? https://github.com/Schniz/fnm/blob/9ba282a705c7a8f3d2dfbfb08d4c62720f600676/src/commands/exec.rs#L23-L29

Would it make sense to add the `using-file` flag back with a deprecation warning?

Schniz:
@alexjeffburke I'm really sorry I didn't remember it. We should have better test coverage!
Thanks for opening the issue. Keep safe

alexjeffburke:
> @alexjeffburke I'm really sorry I didn't remember it. We should have better test coverage!
> Thanks for opening the issue. Stay safe

Oh hey no worries at all - appreciate your jumping on a solution and same to you :)

PR #317, opened 2020-11-02
Bring back --using-file flag with a deprecation warning.

bjornua:
### Change

Reintroduces the `--using-file` flag for the exec command to remain backwards-compatible for a bit

### Problem / Background

The `fnm exec` command after upgrading would break scripts using the `--using-file` flag.

### Refs

Fixes https://github.com/Schniz/fnm/issues/316

### DIFF
--- a/src/commands/exec.rs
+++ b/src/commands/exec.rs
@@ -17,7 +17,7 @@ pub struct Exec {
     #[structopt(long = "using")]
     version: Option<UserVersion>,
     /// Deprecated. This is the default now.
-    #[structopt(long = "using-file", help = "")]
+    #[structopt(long = "using-file", hidden = true)]
     using_file: bool,
 }
 
--- a/src/commands/exec.rs
+++ b/src/commands/exec.rs
@@ -2,8 +2,10 @@ use super::command::Command as Cmd;
 use crate::choose_version_for_user_input::choose_version_for_user_input;
 use crate::choose_version_for_user_input::Error as UserInputError;
 use crate::config::FnmConfig;
+use crate::outln;
 use crate::user_version::UserVersion;
 use crate::version_files::get_user_version_from_file;
+use colored::Colorize;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::process::{Command, Stdio};
 use structopt::StructOpt;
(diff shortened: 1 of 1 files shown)


bjornua:
@Schniz first of all, thanks for coding `fnm` in rust, it is blazingly fast 🏃‍♂️👍.

Would you mind reviewing/considering this change when you have a chance 🙏?

Schniz:
Hey! Thanks for that. I really forgot about it. but I wonder if that's really a bug and we should keep `--using-file`. Maybe it should use the currently used Node if it is not being set? Using the current Node (being declared in `$FNM_MULTISHELL_PATH`) sounds weird because if you have it, you probably have it in your `$PATH` as well, which leads me to think we should have the current implementation (and therefore we should merge this PR) instead of implementing it the previous way.

What do you think @bjornua?

cc @alexjeffburke

bjornua:
> but I wonder if that's really a bug and we should keep `--using-file`. Maybe it should use the currently used Node if it is not being set? Using the current Node (being declared in `$FNM_MULTISHELL_PATH`) sounds weird because if you have it, you probably have it in your `$PATH` as well, which leads me to think we should have the current implementation (and therefore we should merge this PR) instead of implementing it the previous way.

We're using `fnm exec` in the `launch.json` that's used by vscode. Visual studio doesn't always have the necessary environment available depending on how it was launched. Even if it has the environment variables, when switching to a project that uses a different node version, node version is not updated. So it was useful for use to have `fnm` just always respect the `.nvmrc` file with `--using-file`. This no longer seem to be an issue in the new version of `fnm`.

So the remaining issue is that our current scripts that uses `--using-file` will break when developers upgrade `fnm` to latest. We will have developers running both the new and old version of `fnm` until everyone is upgraded. When everyone has upgraded, we can remove the flag from our scripts. So I was hoping there could be a small deprecation period while people upgrade.

Does this make sense?

bjornua:
> but I wonder if that's really a bug and we should keep --using-file. Maybe it should use the currently used Node if it is not being set? Using the current Node (being declared in $FNM_MULTISHELL_PATH) sounds weird because if you have it, you probably have it in your $PATH as well, which leads me to think we should have the current implementation (and therefore we should merge this PR) instead of implementing it the previous way.

Yea, I think `fnm exec` is useful to us when we want to make sure that `.nvmrc` is used to execute something instead of whatever is in the developers environment variables. Otherwise we could just call `node`. I wondering if introducing `fnm exec --using-env` or something could be implemented to cover that use case if it comes up.

Schniz:
Yeah I understand. Let's go with it, since it is deprecated right now but we can un-deprecate (is that a word?) it and reimplement if needed.

Thanks for the contribution!

Issue #320, opened 2020-11-05
`--use-on-cd` doesn't Work if Terminal Opens in Directory and doesn't Switch Back to Default


tsndr:
Some editors do have an integrated terminal which opens directly in the project folder. In that case the node version isn't selected, so I came up with my own solution, but it would be nice to support this scenario out of the box. My solution also fixes that it doesen't switch back to the default version once I leave the project directory.

## My solution `.bash_profile`
It's not the pretties but it works 😅
```bash
eval "$(fnm env)"
[ -s "/usr/local/opt/fnm/etc/bash_completion.d/fnm" ] && . "/usr/local/opt/fnm/etc/bash_completion.d/fnm"
_fnmcd() {
    cd $@
    (([ -s .node-version ] || [ -s .nvmrc ]) && fnm use || fnm use default) &> /dev/null
}
alias cd=_fnmcd
_fnmcd .
```

Schniz:
Thanks! I'm using Zsh so it's done automatically using the shell hooks.
So the change here is the `_fnmcd .`  after declaring the bash function, right?

Regarding:

> My solution also fixes that it doesen't switch back to the default version once I leave the project directory

I don't really understand the issue. can you please elaborate?

tsndr:
## Current Behavior
This is `fmn`'s out of the box behavior.

`.bash_profile`
```bash
eval "$(fnm env --use-on-cd)"
```

### 1. Scenario: Editor
I open my editors terminal, which starts in my project directory so the `--use-on-cd` hook never runs.

### 2. Scenario: Leaving project directory
If I leave the project directory using `cd ..` `fnm` doesen't switch back to my default nodejs version.

## Expected Behavior
The behavoir I'd expect from `fnm` and what my solution fixes.

### 1. Scenario: Editor
I open my editors terminal, therfore my `.bash_profile` is executed and `_fnmcd .` runs my function which checkes if there's a `.node-version` or `.nvmrc` file and runs `fnm use`.

### 2. Scenario: Leaving project directory
Once I leave the project directiory using `cd ..` and since I set an alias to overwrite the `cd` command, it checks again for a `.node-version` or `.nvmrc` file and since it doesn't exist in the current directory it runs `fnm use default` instead of `fnm run` so it switches me back to my default nodejs version.

Schniz:
Regarding 1 — a fix is on its way. It has been like that for ever in Zsh (which I use), but Bash is using aliases and not hooks so it will be fixed in the next version 👍 
[...]

Issue #144, opened 2019-09-30
Better Support for Automatic Switching


CWSpear:
(Sorry, I edited this post quite a bit from the original cuz it was poorly laid out, sorry if this version doesn't match what you might have gotten in an email.)

-----

The `nvm` docs have [scripts](https://github.com/nvm-sh/nvm#zsh) that you can add to your .zshrc, etc, similar to the script used with `fnm env --use-on-cd`, but with better support in searching ancestors for `.nvmrc` and not switching unless the version is actually from the version.

While `fnm` is _much_ faster, the latter is not really _needed_, it'd still be nice to have this sort of capability so that one does not need to see `Using latest-dubnium` every since time they switch directories when it's the same `.nvmrc` (which would happen a lot if it had the parent search).

The `fnm env --use-on-cd` one also doesn't support switching back to `default`.

`nvm` has a couple of things to help out with accomplishing this:

### `nvm version`

`nvm` has a `version` subcommand that will return the exact (installed) version that matches given input, i.e.:

```bash
$ nvm version 6
v6.17.0

$ nvm version lts/dubnium
v10.16.3
```

This does not do a lookup, but it based on aliases, so if `v.10.16.0` is the latest lts/dubnium you have, `nvm version lts/dubnium` will return that.

### `nvm_find_nvmrc`*

This one isn't as big of a deal, because there are lots of other simple scripts out there to do this thing, but it provides a helper function to find the nearest `.nvmrc` by first looking in `pwd` and then checking each ancestor directory.

*I think `nvm_find_nvmrc` must be a zsh-specific thing because it's not used in the [bash version in the `nvm` docs](https://github.com/nvm-sh/nvm#bash). It just uses a fairly standard implementation of "find-up":

```bash
find-up () {
    path=$(pwd)
    while [[ "$path" != "" && ! -e "$path/$1" ]]; do
        path=${path%/*}
    done
    echo "$path"
}
```

Schniz:
> (Sorry, I edited this post quite a bit from the original cuz it was poorly laid out, sorry if this version doesn't match what you might have gotten in an email.)

hehe, no worries, it's a holiday here and I forgot to bring a charger home 😅 


> While `fnm` is _much_ faster, the latter is not really _needed_, it'd still be nice to have this sort of capability so that one does not need to see `Using latest-dubnium` every since time they switch directories when it's the same `.nvmrc` (which would happen a lot if it had the parent search).

[...]

PR #321, opened 2020-11-05
Switching back to default if no `.node-version` or `.nvmrc` file is present

tsndr:
This would be the solution to no. 2 of this #320

### DIFF
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -24,7 +24,7 @@ impl Shell for Bash {
                 __fnmcd () {
                     cd "$@"
                     
-                    if [ -s $(pwd)/.node-version ] || [ -s $(pwd)/.nvmrc ]; then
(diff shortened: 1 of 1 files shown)


Schniz:
Hi, and thanks for your contribution.

As I mentioned in https://github.com/Schniz/fnm/issues/320#issuecomment-722433015, I think that falling back to the default version if we can't find a version file can cause issues. Therefore, at least for now, I think we shouldn't merge this. I think going towards #144 is a better idea. Let's revisit it in the future.

If that change fits your workflow, feel free to avoid `--use-on-cd` and keep your custom script 😄 the `--use-on-cd` script is not the core of fnm anyway, but a nice helper that I use regularly.

I'm sorry that I'm closing it, and I really appreciate your contribution and hope we can collaborate in the future.

tsndr:
No hard feelings 😉

Issue #305, opened 2020-10-27
mksh: thread 'main' panicked at 'Can't infer shell'


saltymouse:
I'm getting the same error using [mksh](https://github.com/MirBSD/mksh) shell even with the new fnm release. Is there a possible similar fix for this shell?

```
thread 'main' panicked at 'Can't infer shell', src/shell/mod.rs:28:32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Related: #296 #290

macOS 10.15.7
mksh: stable 59b
fnm 1.22.3

Schniz:
Hi @saltymouse and thanks for opening the issue!

Unfortunately, `mksh` is not officially supported right now.

You can add the `--shell` option explicitly to tell fnm to produce syntax that matches a different shell, like Bash:

```bash
eval "$(fnm env --shell=bash --use-on-cd)"
```

I verified it locally and it works. I'm open to a PR to make it officially supported, in which we can run all the e2e tests against it too.

From my local tests, looks like `fnm env --shell=bash` will work. I saw in the docs that mksh has hooks (which I don't understand how to use from the docs) and we can maybe find "the mksh way" to implement `--use-on-cd`, instead of overriding `cd` like we do in Bash's solution.

That being said, panicking is not a good way of saying that, so I'll work on a nicer error message.

Would you mind renaming this issue to `Add official support for mksh`?

saltymouse:
Thanks for the info and the workaround. I'm not too familiar with the internals of mksh either, though I can ask the dev. Where should I look in the code to reference making support for this new shell?

Schniz:
* Here is the Bash support for `fnm env`: (which generates the commands in `fnm env`) https://github.com/Schniz/fnm/blob/master/src/shell/bash.rs
* Shell inference [for Unix](https://github.com/Schniz/fnm/blob/master/src/shell/infer/unix.rs) and [for Windows](https://github.com/Schniz/fnm/blob/master/src/shell/infer/windows.rs) (maybe we can have a single function that consolidates this logic that is duplicated in two files, something like `impl From<&str> for Option<Box<dyn Shell>>` implementation or `to_supported_shell(binary_name: &str) -> Option<Box<dyn Shell>>`)
* The [tests/shellcode](https://github.com/Schniz/fnm/tree/master/tests/shellcode) directory has lots of "high-level" shell DSL that should be implemented in order to make the e2e tests (that in `tests/feature_tests`) run against it.

[...]

PR #322, opened 2020-11-05
Run `fnm use` on shell initialization in Bash

### LINKED ISSUE #320: `--use-on-cd` doesn't Work if Terminal Opens in Directory and doesn't Switch Back to Default

Schniz:
This is done in Fish and Zsh already by using their hooks. Bash doesn't does it automatically so this fixes it. Also, it implements what @mirabilos suggested in https://github.com/Schniz/fnm/issues/305#issuecomment-719973203 to make the bash function better.

Fixes #320

### DIFF
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -21,17 +21,19 @@ impl Shell for Bash {
     fn use_on_cd(&self, _config: &crate::config::FnmConfig) -> String {
         indoc!(
             r#"
-                __fnmcd () {
-                    cd "$@"
-
-                    if [[ -f .node-version && .node-version ]]; then
-                        fnm use
-                    elif [[ -f .nvmrc && .nvmrc ]]; then
+                __fnm_use_if_file_found() {
+                    if [[ -f .node-version || -f .nvmrc ]]; then
                         fnm use
                     fi
                 }
 
+                __fnmcd () {
+                    cd "$@" || return $?
+                    __fnm_use_if_file_found
+                }
+
                 alias cd=__fnmcd
+                __fnm_use_if_file_found
             "#
         )
         .into()
--- a/src/shell/fish.rs
+++ b/src/shell/fish.rs
@@ -23,9 +23,7 @@ impl Shell for Fish {
             r#"
                 function _fnm_autoload_hook --on-variable PWD --description 'Change Node version on directory change'
                     status --is-command-substitution; and return
-                    if test -f .node-version
-                        fnm use
-                    else if test -f .nvmrc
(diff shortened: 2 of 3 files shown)


tsndr:
I'd recommend using `-s` instead if `-f`

`-f`: checks if the file exists
`-s` checks if the file exists and is larger than 0 bytes

Schniz:
> I'd recommend using `-s` instead if `-f`
> 
> `-f`: checks if the file exists
> `-s` checks if the file exists and is larger than 0 bytes

Interesting! Didn't know that!

I wonder what's better — to skip the files and fail silently, or let it error out loudly. I like seeing errors so this is why I used `-f` and not `-r` to begin with.

What do you think?

tsndr:
I think I wouldn't necessalily throw a big red error message. I'd just return an info that no file was found.

And I'd also add a `-q` parameter to prevent output, but that's just me 😅

mirabilos:
Tobias Schneider dixit:

>I'd recommend using `-s` instead if `-f`

If that’s desired… I started with the currently-existing code.


Gal Schlezinger dixit:

>https://github.com/Schniz/fnm/pull/322.diff

The “bash” part looks _almost_ fine for all Korn shell variants:
do _not_ place a space before the () for POSIX-style function
definitions; this is a bashism and not parsed by some traditional
Korn shells (mksh implements it, but only because too many people
get it wrong).

Besides that, it reintroduces the fragility problem. *Inside* the
__fnmcd() function definition, use “\cd” instead of just “cd”, to
avoid alias expansion.

bye,
//mirabilos
-- 
FWIW, I'm quite impressed with mksh interactively. I thought it was much
*much* more bare bones. But it turns out it beats the living hell out of
ksh93 in that respect. I'd even consider it for my daily use if I hadn't
wasted half my life on my zsh setup. :-) -- Frank Terbeck in #!/bin/mksh

PR #347, opened 2020-12-04
Check if $ZDOTDIR exists to find correct path to .zshrc

thales-maciel:
Zsh provides a way to keep all related files in a separate directory by exporting a ZDOTDIR variable, but the install.sh script doesn't check It. 

This is a minor issue, but may lead to some bad first time experience for some users, since the required lines would be added in a .zshrc file that is never sourced.


Schniz (approved):
woah, never knew that!
sorry for taking too much time to merge this. thanks!!! 👯

Issue #336, opened 2020-11-23
Can't make system's node the default


zimme:
```sh
➜  ~ fnm --version
fnm 1.22.8
➜  ~ fnm default system
error: Can't create symlink for alias: entity not found
➜  ~
```

Schniz:
sounds like a bug! 🐛 
sorry about that. Hope I'll have time to tackle that soon.

juliankrieger:
@Schniz I'd like to try. Any idea where to start?

zimme:
I wonder if it's the function installation_path in this file that's the problem. It seems to return None for Bypassed i.e. when system node is used. As because if that the alias code can't create a symlink as it doesn't get any path for the system node.

Could be wrong as I'm not familiar with the code base and haven't programmed in rust before.

https://www.github.com/Schniz/fnm/tree/master/src%2Fversion.rs

juliankrieger:
Yes, I agree. When passing "system" to fnm, `choose_version_for_user_input` returns `ApplicableVersion { ..., version: Version::Bypassed`.  Later on, `installation_path` returns `None`. A more correct version would return the output of `which node`, no?

zimme:
I thought I saw ApplicableVersion containing path, does that point to the installation path for the system's node when version i bypassed? And could that be used?

juliankrieger:
Weirdly enough, `ApplicationVersion` points to /dev/null/ for me. I kind of wonder why `create_alias` seemingly does the work twice? `ApplicableVersion`'s `path` property points to the same value ("/Users/julian.krieger/.fnm/node-versions/v8.17.0/installation") as `version_dir` in the `create_alias` function. 

Is there any reason we can't just pass ApplicableVersion to create_alias and use its path, @Schniz ?

Schniz:
I guess we will eventually need to do a refactor for the version structs because they kinda getting out of hand. 
the secret sauce of `system` version (or, Bypassed) is that we provide a symlink that wouldn’t resolve into a Node binary path. So `FNM_SYMLINK:/bin` in `$PATH` will practically make it bypass (hence the name internally) the symlink. This way we don’t need to have dynamic links to the current node version. Just a simple hack!

I still need to read whatever else you wrote because I’m afk so I’m sorry for not helping more than that.

juliankrieger:
[...]

PR #366, opened 2021-01-16
WIP: Fix 'fnm default system' not working

juliankrieger:
This fixes the issue mentioned in #336, namely
```
➜  ~ fnm --version
fnm 1.22.8
➜  ~ fnm default system
error: Can't create symlink for alias: entity not found
➜  ~
```

@Schniz, I could only test this locally on my machine and it seems to be working. I'm also not sure if tests are running correctly, since most of them fail weirdly, even when ran on the most recent master branch.

### DIFF
--- a/src/alias.rs
+++ b/src/alias.rs
@@ -14,14 +14,18 @@ pub fn create_alias(
 
     let version_dir = version
         .installation_path(config)
-        .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::NotFound))?;
+        .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::NotFound));
     let alias_dir = aliases_dir.join(common_name);
 
     if alias_dir.exists() {
         remove_symlink_dir(&alias_dir)?;
     }
 
-    symlink_dir(&version_dir, &alias_dir)?;
+    if *version != Version::Bypassed {
+        let version_dir_res = version_dir?;
+        // Symlink only if version should not bypass system node
+        symlink_dir(&version_dir_res, &alias_dir)?;
+    }
 
     Ok(())
 }


src/alias.rs
```diff
@@ -14,14 +14,18 @@ pub fn create_alias(
 
     let version_dir = version
         .installation_path(config)
-        .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::NotFound))?;
+        .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::NotFound));
```
Schniz, line 17: Can we move this block to be at line 25, so declaration and usage will be very close?

Schniz (commented):
Hey, thanks for the contribution!
Really appreciate doing that.

I'm not entirely sure that this change fixes whatever you said. Look like it swallows the error instead of being direct. It's not supporting aliasing to the system version.

I agree that we could have a nicer error message that says something like "Hey champ, we don't support aliasing to the system version". I think it will be a better fit because once we will support aliasing to a system version.

Since `create_alias` returns `std::io::Result<()>` and can't declare a custom error message without making a big change in the codebase, we can go up the call stack, up to `use.rs`, where we can use the [`ensure!`](https://docs.rs/snafu/0.6.10/snafu/macro.ensure.html) macro to have a pretty printed error and fail before we even get to `create_alias`. Similar to what we do here:

https://github.com/Schniz/fnm/blob/2b3d274804bae2f55c5407adc4584edd9930ece2/src/commands/use.rs#L45-L50

https://github.com/Schniz/fnm/blob/2b3d274804bae2f55c5407adc4584edd9930ece2/src/commands/use.rs#L165-L166

Let me know if that's something you find interesting. If you want that we will do it together over Discord/Telegram, that's a possibility. Just need to find time 😃

juliankrieger:
I think you're correct on this one. Aliasing to the system version is not supported and that's correct, however I kind of expect `fnm default system` to set the current node version to `system`. Logically, this behaviour would not happen in `create_alias` of course, however I would not expect it to print an error message regardless. I'll have a look later and see if I can manage moving this behaviour to `use.rs`.

Schniz:
> however I kind of expect `fnm default system` to set the current node version to `system` without printing an error message

But, this PR does not do that. Right? So it's not fixing `fnm default system`, It's an aesthetic change for better error messages which I like. Can you please update the PR so it would be as I mentioned before? When we have time, we can actually fix the issue. But we can start with better error for now 😄

juliankrieger:
> > however I kind of expect `fnm default system` to set the current node version to `system` without printing an error message
> 
> But, this PR does not do that. Right? So it's not fixing `fnm default system`, It's an aesthetic change for better error messages which I like. Can you please update the PR so it would be as I mentioned before? When we have time, we can actually fix the issue. But we can start with better error for now 😄

Yes, I'll see if I find time this weekend

Schniz:
No pressure 😇 have a great weekend

romgrk:
@juliankrieger still working on this?

dalisoft:
So strange, on M1 it works fine but on Intel machines seems not work

PR #368, opened 2021-01-19
Update 'ring' to '0.16.17'

gucheen:
to fix building problem on mac arm64

according to https://github.com/briansmith/ring/issues/1063#issuecomment-732462547


Schniz:
Cool! can we update to 0.16.17?
The comment you linked says:
> You should use 0.16.17 though

gucheen:
@Schniz updated ring to 0.16.7

Issue #382, opened 2021-01-30
`list` command lists up irrelevant files


chalkygames123:
Hi,

I noticed that when there are irrelevant files to `fnm`, the command incorrectly reports them as versions.

For example, with the default install directory, after running this:

```
touch ~/.fnm/node-versions/foo
```

the `list` command will output:

```
% fnm list
* foo
* system
```

It will cause listing `.DS_Store` files that might be created by macOS Finder when you browse the versions directory.

Schniz:
fnm uses the directory structure as its database, so it makes sense that we will list `foo`. That being said, it makes sense to ignore hidden files. Seems like an easy fix — need to add another condition here:

https://github.com/Schniz/fnm/blob/bcbf3b1d08c65c6b23cd92b187179459e7d35be8/src/installed_versions.rs#L9-L11

Would you want to submit a PR? 😃

chalkygames123:
Thank you for the response.

> Would you want to submit a PR? 😃

No, I'd like to leave it to you.

jungaretti:
@Schniz I could tackle this!

jungaretti:
Oops, looks like it's already fixed. I think #384 should have closed this issue.

Schniz:
Yes! Thanks, and sorry.

PR #384, opened 2021-01-31
Do not list hidden directories as installed versions

scadu:
Addresses #382 

I guess it should be sufficient mitigation for now, unless we'd like to tinker with the file properties etc.


Schniz (approved):
LGTM as a quick fix 😃 
thanks!

PR #397, opened 2021-02-19
Fix minor typos

keirog:


Schniz (approved):
Gotta start using a spell checker 😅 thanks!

Issue #402, opened 2021-03-01
fnm on Apple M1 can't find a prebuilt binary on nodejs.org


vikiival:
I'm using 🍏 M1
Installed fnm via `brew install --build-from-source fnm`

When I want to install any node version (like `fnm install 15`) I get following error
```
Can't download the requested version: Can't find version upstream
```

Schniz:
Unfortunately, fnm does not build Node from source, and right now Node has no prebuilt M1-compatible binaries in their downloads page (https://nodejs.org/dist)

Sadly I don't own M1 (yet..? who knows), and there is no CI that allows testing against M1 machines, so I can't really fix that and test my fixes. My idea is to download regular Node builds and run them with Rosetta. But all of this is gibberish to me and buzzwords I heard over the internet because again, I don't own an M1.

I'd be veeeeeeery happy to see a contribution that allows that or even pointing me towards some CI service I can use to get very slow feedback, but at least some feedback, for fixing the issue for users.

pckilgore:
In the interim, would it be possible to expose an option allowing us to select a specific arch?  M1's run node x86 just fine through Rosetta 2, the immediate blocker is that I can't seem to coax fnm into doing that.

iddar:
I solve this issue use a Home brew under `arch -x86_64 ` (aka ibrew) to instal a `x86_64` version to `fnm` 

```sh
alias ibrew='arch -x86_64 /usr/local/bin/brew'
❯ ibrew install --build-from-source fnm
```

Check your package arch with 

```sh
❯ file $(which fnm)
/usr/local/bin/fnm: Mach-O 64-bit executable x86_64

```

Don't is an ideal way ut works for now

pckilgore:
> I solve this issue use a Home brew under `arch -x86_64 ` (aka ibrew) to instal a `x86_64` version to `fnm`

This seems to require a separate version of homebrew, as the ARM64 homebrew refuses to run those commands.

@Schniz happy to help with a PR here if you're OK with exposing an --arch option on install.

Schniz:
Thanks to @pckilgore, fnm v1.24.0 now will install the binaries that can run using Rosetta. So I'm closing this issue. Thanks for reporting! 😃

PR #417, opened 2021-03-27
feat: adds `--arch` option to `fnm install`

pckilgore:
fnm defaults to downloading the Node architecture that matches the
current system, a good default.

This commits adds an option to override that default where emulating
other architectures is possible and desired, to wit: MacOS 11 is arm64
but is capable of running x86_64 binaries, and developers may desire
running older versions of Node that are unavailble in arm64 form.

works around #402

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -1,0 +1,80 @@
+use crate::system_info::platform_arch;
+
+#[derive(Clone, Debug)]
+pub enum Arch {
+    X86,
+    X64,
+    Arm64,
+    Armv7l,
+    Ppc64le,
+    Ppc64,
+    S390x,
+}
+
+#[derive(Debug)]
+pub struct ArchError {
+    details: String,
+}
+
+pub fn get_default() -> Arch {
+    match from_str(platform_arch()) {
+        Ok(arch) => arch,
+        Err(e) => panic!(e.details),
+    }
+}
+
+fn from_str(s: &str) -> Result<Arch, ArchError> {
+    match s {
+        "x86" => Ok(Arch::X86),
+        "x64" => Ok(Arch::X64),
+        "arm64" => Ok(Arch::Arm64),
+        "armv7l" => Ok(Arch::Armv7l),
+        "ppc64le" => Ok(Arch::Ppc64le),
+        "ppc64" => Ok(Arch::Ppc64),
+        "s390x" => Ok(Arch::S390x),
+        unknown => Err(ArchError::new(&format!("Unknown Arch: {}", unknown))),
+    }
+}
+
+impl ArchError {
+    fn new(msg: &str) -> ArchError {
+        ArchError {
+            details: msg.to_string(),
+        }
+    }
+}
+
+impl std::fmt::Display for ArchError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.details)
+    }
+}
+
+impl std::error::Error for ArchError {
+    fn description(&self) -> &str {
+        &self.details
+    }
+}
+
+impl std::str::FromStr for Arch {
+    type Err = ArchError;
+    fn from_str(s: &str) -> Result<Arch, Self::Err> {
+        from_str(s)
+    }
+}
+
+impl std::fmt::Display for Arch {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let arch_str = match self {
+            Arch::X86 => String::from("x86"),
+            Arch::X64 => String::from("x64"),
+            Arch::Arm64 => String::from("arm64"),
+            Arch::Armv7l => String::from("armv7l"),
+            Arch::Ppc64le => String::from("ppc64le"),
+            Arch::Ppc64 => String::from("ppc64"),
+            Arch::S390x => String::from("s390x"),
+        };
+
+        write!(f, "{}", arch_str)
+    }
+}
--- a/src/downloader.rs
+++ b/src/downloader.rs
@@ -1,3 +1,4 @@
+use crate::arch::Arch;
 use crate::archive;
 use crate::archive::{Error as ExtractError, Extract};
 use crate::directory_portal::DirectoryPortal;
@@ -22,7 +23,7 @@ pub enum Error {
     },
     #[snafu(display("The downloaded archive is empty"))]
     TarIsEmpty,
-    #[snafu(display("Can't find version upstream"))]
+    #[snafu(display("Can't find version/arch upstream"))]
     VersionNotFound,
     #[snafu(display("Version already installed at {:?}", path))]
     VersionAlreadyInstalled {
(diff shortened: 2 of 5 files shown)


pckilgore:
All green except label I lack permissions to add, lmk what else is needed.

Schniz:
Thanks for this! Can we change this to be inside the FnmConfig struct? This way it can be set globally in `fnm env` and will work with `--use-on-cd`

Also, do we really want to set it with a configuration? I think that ideally we can solve it out of the box for M1: download M1 build if exists, or default to use Rosetta. Right now we can always use Rosetta. Does it make sense?

Schniz:
I haven't used any emojis but I am really excited for this. Thank you very much for the amazing effort 😃😃😃

pckilgore:
> Also, do we really want to set it with a configuration? I think that ideally we can solve it out of the box for M1: download M1 build if exists, or default to use Rosetta. Right now we can always use Rosetta. Does it make sense?

I actually prototype-d it this way, but my issue (and I assume I'm not the only one that will hit this with M1 specifically), is that even with an arm build of Node available, the underlying npm package breaks (it detects arm and fails when trying to figure out what to do with a binary).  With a x86 Node, those packages all *just work*, because they perceive their environment as x86 too, and then Rosetta2 makes everything work down the whole stack.

> Thanks for this! Can we change this to be inside the FnmConfig struct? This way it can be set globally in fnm env and will work with --use-on-cd

Yeah that makes more sense, will do.

pckilgore:
2897e9b..e059c8a implements defaulting to x64 / Rosetta for the (dawin, arm64) platform tuple.

I'm not a huge fan, since it may result in old binaries defaulting to x64 versions of Node long after arm64 is available.  But definitely better DX in the short term as fnm should now "just work" on M1.

pckilgore:
Back to green. I'm using the release compiled binary in my path locally (M1), so far so good with my normal direnv + fnm assisted workflow between projects.  Will check back in if I notice anything wonky.

src/arch.rs
```diff
@@ -0,0 +1,95 @@
...
+pub fn default_str() -> &'static str {
+    return crate::system_info::platform_arch();
+}
+
+pub fn get_default() -> Arch {
+    match from_str(default_str()) {
+        Ok(arch) => arch,
+        Err(e) => panic!("{}", e.details),
+    }
+}
```
Schniz, line 35: This can get more Rusty by `impl Default for Arch`. When doing this, we can also `impl ToString for Arch` for `default_str` and then we can let [structopt infer the default value by not giving a value to `default_value`](https://docs.rs/structopt/0.3.21/structopt/#default-values)

src/arch.rs
```diff
@@ -0,0 +1,95 @@
...
+        "x86" => Ok(Arch::X86),
+        "x64" => Ok(Arch::X64),
+        "arm64" => Ok(Arch::Arm64),
+        "armv7l" => Ok(Arch::Armv7l),
+        "ppc64le" => Ok(Arch::Ppc64le),
+        "ppc64" => Ok(Arch::Ppc64),
+        "s390x" => Ok(Arch::S390x),
+        unknown => Err(ArchError::new(&format!("Unknown Arch: {}", unknown))),
+    }
+}
```
Schniz, line 48: This can go into the `FromStr` impl block

src/arch.rs
Schniz: I think that by implementing `Display` (like you do in this file later on), `ToString` is automatically implemented as well. So all we need is to implement `Default` for `Arch` and we will get the Structopt inference 😁

src/downloader.rs
```diff
@@ -22,7 +23,7 @@ pub enum Error {
     },
     #[snafu(display("The downloaded archive is empty"))]
     TarIsEmpty,
-    #[snafu(display("Can't find version upstream"))]
+    #[snafu(display("Can't find version/arch upstream"))]
```
Schniz, line 26: Can we print the arch here, because most users won't override it?

Schniz (approved):
So clean 👏 
Can we update the PR title to be "Support Apple M1 by installing Rosetta Node builds"? 😃 

I also commented more stuff, let me know if you want to address them. I don't mind merging and then updating it in the future. This is already awesome that you contributed it, and I don't want to annoy you with more stuff 👏 👯

Schniz:
> Can we update the PR title to be "Support Apple M1 by installing Rosetta Node builds"? 😃

This will work with Homebrew, I hope. If we want to also support the non-homebrew way and add prebuilt binaries for fnm for M1 we should use the Tier 2 arch:
* https://github.com/rust-lang/rust/issues/73908
* https://github.com/shepmaster/rust/blob/silicon/silicon/README.md

src/arch.rs
pckilgore: I had it this way... then needed to parse in two places so extracted... 

But TIL .parse() is magical.  Surprised how well the type inference works!  I feel like Go and OCAML got drunk and and had a baby and called it Rust.

pckilgore: 53e1a4c

src/arch.rs
pckilgore: 0b3d190

src/arch.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg(windows)]
+/// Get a sane default architecture for the platform.
+pub fn default_str() -> &'static str {
+    return crate::system_info::platform_arch();
+}
+
+impl Default for Arch {
+    fn default() -> Arch {
+        match default_str().parse() {
```
pckilgore, line 32: 🤯

src/config.rs
```diff
@@ -40,6 +46,7 @@ impl Default for FnmConfig {
             base_dir: None,
             multishell_path: None,
             log_level: LogLevel::Info,
+            arch: Default::default(),
```
pckilgore, line 49: 🤯 🤯 💥

src/downloader.rs
pckilgore: Give a mouse a cookie...
image

pckilgore: 6638cb3

pckilgore:
> This will work with Homebrew, I hope. If we want to also support the non-homebrew way and add prebuilt binaries for fnm for M1 we should use the Tier 2 arch:

All good points, but I actually think you can skip this work for now, since the script (you have to force it to not use brew) grabs the x86 version which works just fine with Rosetta 2 (and will download x86 versions of Node which will just work fine).

image

pckilgore:
@Schniz Ok local smoke testing with a release build looking good on my end, feel free to merge away!

Schniz:
Hi, I injured myself in the eye playing basketball. So it'll take time for me to re-review and merge. Sorry about that but I really appreciate your work 😃😃😁

pckilgore:
Yikes, no rush on my part, get better!

idkjs:
Looks like the command may have changed. I tried the above script and based on the response, this is the command that worked:

" arch -arm64 brew install --build-from-source fnm"

idkjs:
> @idkjs it's unclear what you're referring to? Might want to make a separate issue if you need support or clarify your post and how it relates to the PR

Sorry thought I was on the issue which referred to this one.

Schniz (approved):
lgtm 👯 
thanks for this. I will be adding the `ARCH` env var as an output to `fnm env` because I forgot to ask it from you. Anyway, great contribution! 👏 👏 👏

idkjs:
Thank you, Mr. @pckilgore.

pckilgore:
@Schniz Thanks for the guidance!

PR #436, opened 2021-04-21
feat: use arm for aarch64-darwin-node@16 platforms

pckilgore:
Small diff, but way gnarlier conceptually (at least to me) since we weave the arch through then override it.  Happy for suggestions.

Works though: https://twitter.com/0xPCK/status/1384649113795538951

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -1,4 +1,6 @@
-#[derive(Clone, Debug)]
+use crate::version::Version;
+
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub enum Arch {
     X86,
     X64,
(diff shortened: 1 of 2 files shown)


Schniz (approved):
Makes sense, looks like the best way to do that 🤷 
Thank you for doing that! 🙇

src/arch.rs
```diff
@@ -10,26 +12,25 @@ pub enum Arch {
...
+pub fn get_safe_arch<'a>(arch: &'a Arch, version: &Version) -> &'a Arch {
     use crate::system_info::{platform_arch, platform_name};
 
-    // TODO: Handle (arch, name, version) when Node v15+ supports darwin-arm64
-    match (platform_name(), platform_arch()) {
-        ("darwin", "arm64") => "x64",
-        (_, arch) => arch,
-    }
+    return match (platform_name(), platform_arch(), version) {
+        ("darwin", "arm64", Version::Semver(v)) if v.major < 16 => &Arch::X64,
```
pckilgore, line 20: Technically clobbers `--arch` options for darwin/arm.

I suppose we can early return for `Arch::Default.default() != arch`

pckilgore: Although, I'd be interested in hearing a use case for that!...

Schniz: I think the only use case for that would be to allow custom archs, like `musl`, which was requested here: #239

Schniz:
thanks 🤩

PR #453, opened 2021-05-23
Rename duplicate entry of `fnm list` to `fnm list-remote` in documentation

binhonglee:

PR #461, opened 2021-05-29
Fix missing list-remote in command doc

binhonglee:
As mentioned in #453, this fixes the generation script instead of just the doc itself.

cc: @Schniz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+node_modules/
 docs/screen_recording
 /benchmarks/results
 /target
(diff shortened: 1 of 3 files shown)


Schniz (approved):
Amazing!
thank you for the contribution 🙏

.gitignore
```diff
@@ -1,3 +1,4 @@
+node_modules/
```
Schniz, line 1: this is on my global ~/.gitignore 😃 
nice addition

Issue #314, opened 2020-10-31
Use behind a corporate proxy


XiaofengXie16:
Is there an option where I could add a certificate to use this tool given that I want to use this tool behind a corporate proxy?

Currently, I am seeing this error.
HttpError: error sending request for url (https://nodejs.org/dist/index.json): error trying to connect: invalid certificate: UnknownIssuer

Schniz:
Hi, I'm going to need to know more details. What OS are you using? what version of fnm?
Thanks

skyrex-mark:
10.15: "Catalina"
fnm version: 1.22.6
Note: NVM works fine behind the proxy

Schniz:
Can you try building fnm on your machine and tell me if it is working well?
I wonder if I should not use `rustls` on Mac 🤔 maybe use it only on ARM

zicklag:
I'm getting a possibly related issue when running behind a corporate proxy with authentication and a custom certificate ( which I have trusted on my machine and which works will most other apps such as `curl` ). I'm on Pop!_OS ( Ubuntu ) 20.04 and I get the following error:

```
[2020-11-23T21:10:45Z ERROR rustls::session] TLS alert received: Message {
        typ: Alert,
        version: TLSv1_2,
        payload: Alert(
            AlertMessagePayload {
                level: Fatal,
                description: HandshakeFailure,
            },
        ),
    }
error: HttpError: error sending request for url (https://nodejs.org/dist/index.json): error trying to connect: received fatal alert: HandshakeFailure
```

zicklag:
Ah, I was able to fix it by disabling the `rustls-tls` feature in `reqwests` and by re-enabling the default features ( which just enables the `default-tls` feature ). It's probably due to RusTLS failing to trust the system certificate.

Schniz:
Oh cool, thanks for checking it out! 
we probably need to use `rustls-tls` only on ARM 😄

pfiaux:
I ran into a similar issue, I'll need to retest with the latest release which has the newest `reqwests` version to make sure it still happens.

Looking at their doc tho it seems `rustls-tls` is equivalent to `rustls-tls-webpki-roots` (https://docs.rs/reqwest/0.11.3/reqwest/#optional-features). Can it be combined with `native-tls` or perhaps `rustls-tls-native-roots`? If they're not exclusive that might allow using the configured certificates on the host as well.

pfiaux:
[...]

PR #468, opened 2021-06-05
Include feature for (native) certificates

pfiaux:
Addresses https://github.com/Schniz/fnm/issues/314 allowing use of fnm when behind a proxy with certificates.

This adds the `"rustls-tls-native-roots"` [cargo feature](https://docs.rs/reqwest/0.11.3/reqwest/#optional-features) to reqwest. That uses [rustls-native-certs](https://github.com/rustls/rustls-native-certs):
> rustls-native-certs allows rustls to use the platform's native certificate store when operating as a TLS client.

I tested on macOS both with and without proxy mainly doing `cargo run -- list-remote`, and installing node 16, this worked behind proxy. Without proxy I only tested list-remote. Perhaps other commands need to be retested as well.

The [`native-tls`](https://docs.rs/native-tls/0.2.7/native_tls/) feature might also work, I didn't test. They seem to cover similar functionality but maybe it works better on Linux ARM systems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ description = "Fast and simple Node.js version manager"
 serde = { version = "1.0.126", features = ["derive"] }
 clap = "2.33.3"
 structopt = "0.3.21"
-reqwest = { version = "0.11.3", features = ["blocking", "json", "rustls-tls", "brotli"], default-features = false }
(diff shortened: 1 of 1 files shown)


Schniz:
Awesome! Thanks. Let's see if all e2e tests pass 🙏

Issue #482, opened 2021-06-24
--install-if-missing does not work fow `lts/*`


Schniz:
this is how it can be reproduced:
```sh-session
$ fnm uninstall 'lts/*'
$ echo 'lts/*' > .node-version
$ fnm use --install-if-missing
```

_Originally posted by @Schniz in https://github.com/Schniz/fnm/issues/144#issuecomment-803909456_

_Reported by @miszo_

Schniz:
Fixed by @AlexMunoz!

PR #484, opened 2021-06-30
fix:  --install-if-missing when using version alias

AlexMunoz:
I am trying to fix this issue https://github.com/Schniz/fnm/issues/482 based on my basic Rust knowledge. Any advice  to improve this further would be appreciated. If I am completely wrong, please let me know that as well.

### DIFF
--- a/src/commands/use.rs
+++ b/src/commands/use.rs
@@ -42,14 +42,13 @@ impl Command for Use {
             system_version::path()
         } else if let Some(alias_name) = requested_version.alias_name() {
             let alias_path = config.aliases_dir().join(&alias_name);
-            ensure!(
-                alias_path.exists(),
-                CantFindVersion {
-                    version: requested_version
-                }
-            );
-            outln!(config#Info, "Using Node for alias {}", alias_name.cyan());
-            alias_path
+            if alias_path.exists() {
+                outln!(config#Info, "Using Node for alias {}", alias_name.cyan());
+                alias_path
+            } else {
+                install_new_version(requested_version, config, self.install_if_missing)?;
+                return Ok(());
+            }
         } else {
             let current_version = requested_version.to_version(&all_versions, &config);
             match current_version {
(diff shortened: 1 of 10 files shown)


src/commands/use.rs
```diff
@@ -19,6 +19,29 @@ pub struct Use {
...
+
+        Self {
+            version: Some(UserVersionReader::Direct(requested_version)),
+            install_if_missing,
+        }
+        .apply(config)?;
+
+        return Ok(());
+    }
+}
```
Schniz, line 43: can we move this out of the `impl Use`?

Dangling functions are okay -- `fn install_new_version` is all good. We can also move the `ensure` block inside to share it among the two branches.

Schniz (commented):
YES! Very good! Thank you.
Can you add a test or at least verify that it works?
Also please look at my comments 🙏

AlexMunoz:
Hi @Schniz  I have made some changes and also added some tests!

Schniz:
Thanks! Eager to see. But I'm afk this weekend. Will check on Sunday! 😃😃😃

src/commands/use.rs
```diff
@@ -61,26 +60,9 @@ impl Command for Use {
...
-                    .context(InstallError)?;
-
-                    Self {
-                        version: Some(UserVersionReader::Direct(requested_version)),
-                        install_if_missing: self.install_if_missing,
-                    }
-                    .apply(config)?;
-
+                    let can_install =
+                        self.install_if_missing || should_install_interactively(&requested_version);
```
Schniz, line 64: can we move the `should_install_interactively` to be a part of the extracted function?
this is for suggesting the user `hey, this is missing, wanna install this?` -- and not just waiting for the `--install-if-missing` flag :)

Schniz: Other than that, it looks reaaaaaaally good. I hope it was easy figuring out how to write the e2e tests :)

AlexMunoz: I made the change ;)  Yes, it took me a while to figure out how to create the snapshots. :D

Schniz (approved):
Amazing!

AlexMunoz:
don't forget the PR label :smile:

PR #510, opened 2021-08-09
feat: add `--reinstall-packages-from=` argument for fnm install

libook:
I am a noob in Rust.
My code may not be good enough. Please feel free to left your comments.

### DIFF
--- a/src/commands/install.rs
+++ b/src/commands/install.rs
@@ -15,6 +15,10 @@ use structopt::StructOpt;
 
 #[derive(StructOpt, Debug, Default)]
 pub struct Install {
+    /// Install global packages from an existing version
+    #[structopt(long)]
+    pub reinstall_packages_from: Option<UserVersion>,
+
     /// A version string. Can be a partial semver or a LTS version name by the format lts/NAME
     pub version: Option<UserVersion>,
 
@@ -24,17 +28,20 @@ pub struct Install {
 }
 
 impl Install {
-    fn version(self) -> Result<Option<UserVersion>, Error> {
-        match self {
+    fn version(&self) -> Result<Option<UserVersion>, Error> {
+        match &self {
             Self {
+                reinstall_packages_from: _,
                 version: Some(_),
                 lts: true,
             } => Err(Error::TooManyVersionsProvided),
             Self {
+                reinstall_packages_from: _,
                 version: v,
                 lts: false,
-            } => Ok(v),
+            } => Ok(v.clone()),
             Self {
+                reinstall_packages_from: _,
                 version: None,
                 lts: true,
             } => Ok(Some(UserVersion::Full(Version::Lts(LtsType::Latest)))),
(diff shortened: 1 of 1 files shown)


src/commands/install.rs
```diff
@@ -98,6 +105,38 @@ impl super::command::Command for Install {
...
+                        "| command sed 1,1d ",
+                        "| command sed ",
+                        "-e '/ -> / d' ",
+                        "-e '/\\(empty\\)/ d' ",
+                        "-e 's/^.* \\(.*@[^ ]*\\).*/\\1/' ",
+                        "-e '/^npm@[^ ]*.*$/ d'",
+                        ),
+                        user_version,
+                    ))
+                    .output().expect("Failed to get package list.");
```
Schniz, line 128: This will not work on Windows-based shells (wincmd).
Maybe we can try adding a new command, `fnm reinstall-packages-from`, and we can add unit/e2e tests for it.

src/commands/install.rs
```diff
@@ -157,6 +227,10 @@ pub enum Error {
     CantFindRelevantLts {
         lts_type: crate::lts::LtsType,
     },
+    #[snafu(display("Execute command failed:\n{}", stderr))]
+    CommandExecutionFailed {
+        stderr: String,
+    },
```
Schniz, line 233: Please be more specific. `InstallingPackagesFromOtherVersionFailed`

Schniz (commented):
Hi @libook! THANK YOU for the contribution!

As I already mentioned in the thread, I am pretty torn about this feature: it really feels out of fnm's scope. You can even see that it is very third party in the code you just wrote here, and can be easily migrated to be a script like:

```
#!/bin/sh

fnm install $1
PACKAGES=$(fnm exec --using=$2 npm list -g --depth=0 2>/dev/null | ...)
fnm exec --using=$1 npm install -g $PACKAGES
```

So I'm not entirely sold in favor of maintaining this. What do you think? What are the pros (and if you can, the cons) of adding this?

---

Also, since fnm supports Windows, we should support this feature for Windows users too. Taking all of this code to another Rust module that is properly tested using unit/e2e tests will allow CI to run the code on Linux/Mac _and_ Windows machines.

---

thanks again for your contribution and I'm really sorry if I'm a serious downer.

libook:
@Schniz You are right.

I am an nvm user for many years. I use MacOs, Linux and WSL as developing environment. And have forgotten that nvm doesn't support windows.

I am learning Rust. I found fnm in some Rust news. It is a real "fast node manager" as the name said.
I use nvm in many automatic scripts. So I wish that I can replace nvm to fnm. Fnm doesn't have features like `--reinstall-packages-from=`. I was wondering if I can add the feature for fnm.

That is why I did this.

---

> it is very third party in the code

I agree with you. This feature requires both npm and node version manager. Npm can't do this by itself. Extra script is the only way, if a node version manager doesn't do this either.
It may still be a common requirement while using a node version manager (at least for developers who is maintaining multiple node projects like me).

Maybe there will be a better way for cross platform supporting and even for multiple node package managers too(e.g. yarn and pnpm) in the future.

I am happy with learning and hacking Rust in this work. And I have learned a lot. I am ok if you think that it is not the time for this feature.

ljharb:
@libook the next version of nvm does in fact support windows, the current one has supported WSL for years.

libook:
@ljharb Wow, a familiar avatar. I always see you in ECMA works.

Yes, I have been using nvm for years. Nvm has powerful functions. I have a [zsh tool box](https://github.com/libook/IdleBox) which requires nvm as an optional dependency for scripts.
Will nvm support Windows PowerShell? I know nvm-windows. But POSIX environment (sure, include WSL) is enough for me. Just in case for native library requirements you know.

Thank you guys for maintaining cool things.

gatspy:
> 

mac osx zsh run this script , error info `command not found`

PR #514, opened 2021-08-13
Invoke fnm use on startup for PowerShell

### LINKED ISSUE #513: Load .nvmrc / .node-version on shell startup (PowerShell)
Wasn't sure about bumping an old and closed issue, but I'm seeing the same issue as #410 except with PowerShell on Windows 10. The `.nvmrc` is not auto loaded on startup.

naoey:
I believe this should fix the issue and seems to work running it with `cargo run -- env --use-on-cd` as it would be used on the release build.

Fixes #513


Schniz (approved):
Cool! thanks for the PR! 😇😇

PR #516, opened 2021-08-17
More informative log level error message

waldyrious:
Ideally, [the docs](https://github.com/Schniz/fnm/blob/master/docs/commands.md) should mention these as well, and I'd like to include that change in this PR, but I'm not sure how to insert the list into the current documentation format:

        --log-level <log-level>
            The log level of fnm commands [env: FNM_LOGLEVEL]  [default: info]

Any ideas, @Schniz? Maybe something like this?

        --log-level <log-level>
            The log level of fnm commands [env: FNM_LOGLEVEL]  [supported: quiet, info, error; default: info]

Or this, to use the [docopt](http://docopt.org/) syntax?

        --log-level (quiet|info|error)
            The log level of fnm commands [env: FNM_LOGLEVEL]  [default: info]

### DIFF
--- a/src/log_level.rs
+++ b/src/log_level.rs
@@ -44,7 +44,7 @@ impl std::str::FromStr for LogLevel {
             "quiet" => Ok(Self::Quiet),
             "info" | "all" => Ok(Self::Info),
             "error" => Ok(Self::Error),
-            loglevel => Err(format!("I don't know the log level of {:?}", loglevel)),
+            loglevel => Err(format!("Unrecognized log level {:?}. Supported levels are 'quiet', 'info' and 'error'.", loglevel)),
         }
     }
 }
--- a/src/log_level.rs
+++ b/src/log_level.rs
@@ -44,7 +44,10 @@ impl std::str::FromStr for LogLevel {
             "quiet" => Ok(Self::Quiet),
             "info" | "all" => Ok(Self::Info),
             "error" => Ok(Self::Error),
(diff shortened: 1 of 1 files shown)


src/log_level.rs
```diff
@@ -44,7 +44,10 @@ impl std::str::FromStr for LogLevel {
             "quiet" => Ok(Self::Quiet),
             "info" | "all" => Ok(Self::Info),
             "error" => Ok(Self::Error),
-            loglevel => Err(format!("I don't know the log level of {:?}", loglevel)),
+            loglevel => Err(format!(
+                "Unrecognized log level {:?}. Supported levels are 'quiet', 'info' and 'error'.",
```
Schniz, line 48: I wonder if there's a way to automatically extract it. But anyway, I'm merging because it is important fix! thank you!

Schniz (approved):

waldyrious:
@Schniz any thoughts on how the docs could also include this information, as per my suggestions in the opening comment?

Schniz:
@waldyrious looks like we can implement `LogLevel::possible_values` and add `https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.possible_values` to reference to it. Then it should show in the docs 😃 I at least hope so

waldyrious:
Ok, nice. Is that something you're planning to work on, or are you suggesting that I do it? 'Cause if it's the latter case, I will probably need a bit more guidance :)

Schniz:
I will file a new issue with explanations what code should go where 😃

Schniz:
@waldyrious opened #523

PR #519, opened 2021-08-20
Windows: Use junctions rather than symlinks

davidaurelio:
On windows, creating symlinks requires special privileges.

These privileges are typically either assigned by activating windows’s “Developer Mode” or via security policies.

This makes using _fnm_ more difficult for beginners and people on managed machines, where developer mode might not include the right to create symlinks, and local security policies are reset on a regular basis.

NTFS offers an alternative mechanism. Junctions can be **created without special privileges.**

[superuser.com] has some more extensive answers on the differences between directory symlinks and junctions. The relevant difference for fnm is that symlinks can work across network drives, whereas junctions only work locally. This shouldn’t matter for fnm’s functionality.

By switching to directory junctions, fnm becomes easier to work with on windows, especially under circumstances where privileges to create symlinks cannot be assigned permanently.

This change is backwards compatible: Since both directory symlinks and junctions are removed using `std::fs::remove_dir`, an fnm upgrade will keep working with existing setups.

[Directory Junctions]: https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions#junctions
[superuser.com]: https://superuser.com/questions/343074/directory-junction-vs-directory-symbolic-link

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -458,6 +458,7 @@ dependencies = [
  "env_logger",
  "indoc",
  "insta",
+ "junction",
  "log",
  "pretty_assertions",
  "reqwest",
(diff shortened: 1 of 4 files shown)


Schniz:
Woah!!! Sounds amazing. Let's see that the tests pass. 💯💯💯

PR #528, opened 2021-09-10
installation script: Use `=` instead of `==`

develoot:
Posix test does not support `==`. It use `=` for value comparison.

Because of this, it seems some shells (e.g., zsh) does not process the script well.

`==` works only for bash.


Schniz (approved):
Thanks! My inner JavaScript dev already felt like `==` is a compromise hehe. Good catch!

PR #544, opened 2021-10-10
Replace error message with a more useful one

yonifra:


Schniz:
I hope that it will make it easier to understand the issue 😃

PR #547, opened 2021-10-12
Replace reqwest with ureq for less dependencies and smaller file size

dnaka91:
First of all great project, although not mainly a JS/TS developer I use it quite often and really like the ease of use of it and compability with different shells :tada:.

I saw that `fnm` is using `reqwest` with the `blocking` feature enabled, which has a big impact on the dependency count and file size. The reason is that reqwest depends on the `tokio` runtime and enabling the `blocking` feature internally runs a tokio runtime with a non-async interface over the usual API. This leads to many dependencies and by using a different popular HTTP client library (namely `ureq` but there are several others) , both dep count and file size can be well improved.

`ureq` is a non-async HTTP client library and has a very small dependency count. Below are the results with **Before** when using `reqwest` and **After** when using `ureq`:

| Value             | Before | After | Difference |
| ----------------- | ------ | ----- | ---------- |
| Dependencies      | 208    | 151   | -27.4%     |
| File size (MacOS) | 7.7MB  | 4.9MB | -33.8%     |

I added a dependency on `url` as well, as I saw that it's used in many places (re-exported from `reqwest`) and wanted to keep it that way as I thought you might not want to replace all spots with a simple `&str` or `String` as the `Url` type gives you the guarantee of having a valid URL wherever you use it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "35ef4730490ad1c4eae5c4325b2a95f521d023e5c885853ff7aca0a6a1631db3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "697ed7edc0f1711de49ce108c541623a0af97c6c60b2f6e2b65229847ac843c2"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "ansi_term"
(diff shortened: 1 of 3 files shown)


Schniz:
I am sure there are ways to overcome that. We can add a "accept-encoding" header and extract manually

dnaka91:
Okay just added it. There are actually just 2 specific locations where we make HTTP calls.
- For the node distribution
- For the index JSON which contains information about all available distributions.

First I thought it's not necessary but I saw the JSON is about 250kb uncompressed and probably growing in the future. So for the index file only I added the brotli decompression. Not needed for the distributions themselves as they're already packed in an archive, I would say.

Schniz:
@dnaka91 I'm rerunning the entire workflow. GitHub Actions was down for Windows runners: https://twitter.com/githubstatus/status/1448242741067005952

PR #554, opened 2021-10-13
Fix clippy & use musl target on Rust compiler for static compilation

dnaka91:
Just some cleanups here and there, mostly related to clippy:

- Fix all existing clippy lints reports for the project.
- Additionally enable `clippy::pedantic` and `rust_2018_idioms` lints and fix any reported warnings.
- Update CI to include an extra step for running clippy.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -77,7 +77,7 @@ jobs:
         path: target/release/fnm.exe
 
   build_macos_release:
-    runs-on: macOS-latest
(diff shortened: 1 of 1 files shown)


Schniz:
> You want to have a fully statically compiled binary right?

Yup, you are right 😃 I don't remember why I needed to use a container. Maybe it's just something I did the same as I did on the ReasonML version

Schniz:
Amazing! thank you!

PR #556, opened 2021-10-14
Allow aliasing to the system version

### LINKED ISSUE #439: `fnm default system` does not work
Hi @Schniz.

It's again me, thank you for developing this amazing tool.

My system now updated to Node.js v16 and i tried fnm to use Node.js v14 and it worked fine.
But reverting does not work `fnm use system`.

## Environments
- OS: macOS Big Sur 11.2.3
- Arch: **Apple Silicon** and **x64 Intel**
- Shell: fish
- Prompt: starship
- Node.js: System v16, calling v14

## Related issue

> Maybe it's not related

- #381 
- #366 
- #336

Schniz:
this PR allows `fnm alias system any_alias` and specifically `fnm default system`, for those who are seeking for it.

it also reveals that we need to do some housekeeping and clean the code. it is very complicated and I believe I can make it nicer. but this is for the future.

This fixes #439

### DIFF
--- a/src/alias.rs
+++ b/src/alias.rs
@@ -13,9 +13,7 @@ pub fn create_alias(
     let aliases_dir = config.aliases_dir();
     std::fs::create_dir_all(&aliases_dir)?;
 
-    let version_dir = version
-        .installation_path(config)
-        .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::NotFound))?;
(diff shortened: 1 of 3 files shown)


Schniz:
I will try to use Vagrant & VSCode remote for understanding what's wrong with the Windows tests.
@github we need Windows hosts in Codespaces for open source! 😁

PR #565, opened 2021-10-27
Replace wmic with PowerShell

### LINKED ISSUE #561: fnm env produce error: Can't infer shell!
Recently im facing this error,
```
error: Can't infer shell!
fnm can't infer your shell based on the process tree.
Maybe it is unsupported? we support the following shells:
* cmd
* powershell
* bash
* zsh
* fish
```
Im using windows 11 , and before im not facing this issue,
i did try to reinstall fnm and build from cargo
but the error persist.

is there a way to fix this?
Only in windows this is persisting, im using it in all my dev machine, artix linux, mac big sur no problem.

hope this will be resolve
coz the command wont simple generate the correct fnm path for windows now.
so the command node and npm are both not found.

i just use pure node js installation for now until this is resolve 
thanks for building fnm

dabide:
Fixes #561

### DIFF
--- a/src/shell/infer/windows.rs
+++ b/src/shell/infer/windows.rs
@@ -6,11 +6,11 @@ use std::collections::HashMap;
 
 #[derive(Deserialize, Debug)]
 pub struct ProcessInfo {
-    #[serde(rename = "ExecutablePath")]
+    #[serde(rename = "Name")]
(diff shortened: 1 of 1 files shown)


Schniz:
Superseded by #573. Sorry that I didn't merge it!

PR #598, opened 2021-12-13
Update README.md file emoji

lorensr:
Feels more positive/friendly to me than the current 🤔. Some other options are:

page_facing_up
open_file_folder
clipboard

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 :rocket: Built with speed in mind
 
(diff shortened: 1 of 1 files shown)


Schniz:
Ahhh I feel you. Let's not use ✅  as it's very basic and makes the other ones look as if they're not implemented. I like :open_file_folder: better 😃

lorensr:
@Schniz done!

PR #603, opened 2021-12-16
(re)Include feature for (native) certificates

pfiaux:
Addresses #314 allowing use of fnm when behind a proxy with certificates by reapplying #468 after the revert from ureq to reqwest.

This adds the `"rustls-tls-native-roots"` [cargo feature](https://docs.rs/reqwest/0.11.3/reqwest/#optional-features) to reqwest. That uses [rustls-native-certs](https://github.com/rustls/rustls-native-certs):
> rustls-native-certs allows rustls to use the platform's native certificate store when operating as a TLS client.

I reused the previous commit message mostly, but not 100% accurate since this I didn't run the same tests this time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -245,6 +245,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


pfiaux:
I wasn't able to clone and build locally yet, I might be able to give it a try next week.

Schniz:
Is it okay that I will push the changes in `Cargo.lock` to this branch?

pfiaux:
Sure! let me know if I need to do something on my side as well.

PR #607, opened 2021-12-20
Allow recursive version lookups

### LINKED ISSUE #144: Better Support for Automatic Switching

### LINKED ISSUE #442: --use-on-cd should revert to default node
The flag `--use-on-cd` should revert to default node if there is no `.nvmrc` present.
Is this already the case? Because it's just staying on the last used node version when switching in directories without one.

Schniz:
Adding the ability to provide `--version-file-strategy=recursive` on
both `fnm env` and on each command can allow us to support recursive
resolutions of the Node versions.

This makes the default Node version as easy as:

```bash
echo "16" > ~/.node-version
```

(assuming your code lives inside your home directory) and fixes #442, if
becomes the default.

---

This fixes #144 which is a very old issue :smiley:
https://github.com/Schniz/fnm/issues/144#issuecomment-770798955

---

This code is under a feature toggle, so we can merge it and consider whether we want to use it eventually.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -579,7 +579,7 @@ FLAGS:
         --install-if-missing
             Install the version if it isn't installed yet
 
-        --silent-when-unchanged
+        --silent-if-unchanged
             Don't output a message identifying the version being used if it will not change due to execution of this
             command
     -V, --version
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -28,6 +28,14 @@ OPTIONS:
         --node-dist-mirror <node-dist-mirror>
             https://nodejs.org/dist/ mirror [env: FNM_NODE_DIST_MIRROR]  [default: https://nodejs.org/dist]
 
+        --version-file-strategy <version-file-strategy>
+            A strategy for how to resolve the Node version. Used whenever `fnm use` or `fnm install` is called without a
+            version, or when `--use-on-cd` is configured on evaluation.
+
+            * `local`: Use the local version of Node defined within the current directory
+
+            * `recursive`: Use the version of Node defined within the current directory and all parent directories [env:
+            FNM_VERSION_FILE_STRATEGY]  [default: local]  [possible values: local, recursive]
 
 SUBCOMMANDS:
     alias          Alias a version to a common name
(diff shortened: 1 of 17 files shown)


Schniz:
One thing I wonder about this solution, is how would one use `fnm use`, if `--use-on-cd` is installed. Maybe they don't? Automatically switching to default version can be annoying

jimmythomson:
@Schniz - thanks very much for adding this feature. I would have stuck with nvm had it not been for this as I'm constantly jumping between monorepos where a single top level `.nvmrc` is defined. Any idea when this functionality will go in to a release?

Schniz:
I will release it tomorrow morning (Israel time). 😊

Issue #475, opened 2021-06-14
Xcode is having trouble finding node


maharjanaman:
When I try to build react-native project on xcode, I get
```
env: node: No such file or directory 
Command PhaseScriptExecution failed with a nonzero exit code
```
On shell(zsh), I don't have any trouble with node and `which node` print outs `/var/folders/nz/hh7gg42x1tjf443r4fq3xylm0000gn/T/fnm_multishells/33995_1623643112968/bin/node`.

Is there anything I'm missing during installation process?

joeyfigaro:
Having same issue. I tried updating RN to make use of their `find-node.sh` script, but it looks like it's looking for `NODE_BINARY`, which is undefined. Is there a recommended way to keep `NODE_BINARY` pointing to the current version used by fnm?

maharjanaman:
@joeyfigaro Was able to run the project by installing another node version using brew i.e. `brew install node`, which then acts as system node.

Schniz:
duplicate of #486 (please reopen if I'm getting this wrong)

douglasjunior:
For `nvm` we can just run `ln -s $(which node) /usr/local/bin/node` to create a link to node binary, but with `fnm` this dont work. 😢

Schniz:
You can run `ln -s "$(realpath $(which node))" /usr/local/bin/node`, but I am not sure I would suggest so.

Example:
```sh-session
$ echo "$(realpath $(which node))"
/Users/<username>/.fnm/node-versions/v14.15.4/installation/bin/node
```

douglasjunior:
Thanks @Schniz this looks good.

And about `husky js`, any suggestions?

I tried to add `export PATH=$HOME/.fnm:$PATH` to the `~/.huskyrc`, but don't work.

taylorkline:
I thought about having an `.xcode.local` with:

```
export NODE_BINARY=/System/Volumes/Data/Users/username/Library/Application Support/fnm/aliases/lts-hydrogen/bin/node
```

This does not work due to the space in `"Application Support"` (even if escaped!)

However, a symlink `ln -s /System/Volumes/Data/Users/username/Library/Application\ Support/fnm/aliases/lts-hydrogen/bin/node ~/node-18` allows for:

```
export NODE_BINARY=/Users/username/node-18
```

To work just fine.

kud:
I've also got this issue with react-native.

For instance, the `.xcode.env` file is:

```
[...]

PR #609, opened 2021-12-21
Add shims for Volta-like binaries

Schniz:
This PR introduces `--with-shims` to the `fnm env` command.
These shims (currently only `node`, but maybe `npm` will need it too) will be used instead of the `$FNM_MULTISHELL_PATH`, and although they will introduce some runtime penalty (~8-16ms on my machine), it will be much easier to integrate with other tools and make people understand how fnm works.
This should also fix some `rehash`ing issues some users have on newer Zsh versions.


Schniz:
d1f7b9d is because I thought I would be able to change the symlink generation process from random to using the parent process' pid, and therefore making it reproducible, and making `fnm env` not mandatory at all (unless one wants fancy features).

Unfortunately, every shell script has its own pid so we need to find a common denominator somehow. Shells are obviously not the common denominator (the shims are a shell script), so we need to know whether it's an interactive shell--or a graphical IDE (like #475 and other issues).

If someone has an idea how to pull this off, let me know 😃

Schniz:
> Shells are obviously not the common denominator (the shims are a shell script)

CLI (pid1) -> node shim (pid2) -> fnm (pid3)

fnm tries to take pid2, instead of pid1. The same can be applied in VSCode or XCode or whatever:

VSCode (pid1) -> some 3rd party script (pid2) -> another 3rd party (pid3) -> fnm (pid4)

and now we're trying to use pid3, which can be a one-off thing. So pids are probably not what we want. I also thought about implementing a table of pids to versions but this is just insanity and composes very very badly (how up the tree we need to update version when we switch?) 😁

Schniz:
The premise of reproducible "multishell" paths is good, because it simplifies the installation of fnm, which can be awesome.

src/shims/unix.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+fnm exec --using-current -- node "$@"
```
alexeyten, line 3: May be you should use `exec fnm exec ...` so fnm's process replaces shell and process tree would be like this:

CLI (pid1) -> ~~node shim (pid2)~~ -> fnm (pid2)

Not sure how to do this in windows.

Schniz: oh! TIL!

this is still the problem because we have unknown parents that can bridge it. :(

shell -> scripts/run-tests.sh -> scripts/run-unit-tests.sh -> fnm exec ...

pids don't solve it 😭

ljharb:
Shims are a fundamentally different model than PATH hijacking, and they come with an entirely distinct set of problems; I’ll be very curious to see how fnm deals with both at once.

alexeyten:
I've read some linked issues and it seems that people want fnm to install node globally (or behave like it's installed globally). Am I correct?

alexeyten:
What about this shim?
```sh
#!/bin/sh

cwd=$PWD

while [ -n "$cwd" ]; do
    if [ -f "$cwd/.node-version" ]; then
        # echo "Found $cwd/.node-version"
        exec fnm exec --using "$cwd/.node-version" -- node "$@"
    fi
    if [ -f "$cwd/.nvmrc" ]; then
        # echo "Found $cwd/.nvmrc"
        exec fnm exec --using "$cwd/.nvmrc" -- node "$@"
    fi
    cwd=${cwd%/*}
done

exec fnm exec --using default -- node "$@"
```

Schniz:
We can have `fnm exec --fallback-to-default` but I am not sure this is enough.. what if someone wants to change a node version?

 I just wonder if there's a better option we haven't though of, that will make `fnm env` optional

alexeyten:
> what if someone wants to change a node version?

You meant beside changing default version?

docs/commands.md
```diff
@@ -194,6 +194,9 @@ FLAGS:
     -V, --version
             Prints version information
 
+        --with-shims
```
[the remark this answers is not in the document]
Schniz: it was never added, I'm gonna add this with a proper warning, I will test it on my machine for now <picture data-single-emoji=":relieved-telegram:" title=":relieved-telegram:">:relieved-telegram:</picture>

Schniz:
main problem I see with it right now, that it will make all global node module installations fail, as they install binaries in `$NODE_INSTALL_DIR/bin`, and we're not linking it to `$PATH` <picture data-single-emoji=":sadblob:" title=":sadblob:">:sadblob:</picture>

gotta dig deeper <picture data-single-emoji=":dig-deep:" title=":dig-deep:">:dig-deep:</picture>

PR #614, opened 2021-12-28
Make installation script respect $XDG_DATA_HOME

jonz94:
Related to #357 and #613


.ci/install.sh
```diff
@@ -2,7 +2,12 @@
 
 set -e
 
-INSTALL_DIR="$HOME/.fnm"
+if [ -d "$HOME/.fnm" ]; then
+  INSTALL_DIR="$HOME/.fnm"
+else
+  INSTALL_DIR="${XDG_DATA_HOME:-$HOME/.local/share}/fnm"
```
Schniz, line 8: I wonder if that's enough. MacOS way would be `~/Library/Application Support/fnm`

what do you think? SHould we keep `~/.local/share/fnm` just for the installation process?

jonz94: Oops, thanks for your friendly reminder.

I just thought on MacOS `fnm` would be installed at `~/.local/share/fnm`

.ci/install.sh
```diff
@@ -2,10 +2,19 @@
 
 set -e
 
-INSTALL_DIR="$HOME/.fnm"
 RELEASE="latest"
 OS="$(uname -s)"
 
+if [ -d "$HOME/.fnm" ]; then
+  INSTALL_DIR="$HOME/.fnm"
+elif [ ! -z $XDG_DATA_HOME ]; then
```
page-down, line 10: You can use `test -n` and double quotes.
```bash
elif [ -n "$XDG_DATA_HOME" ]; then
```

.ci/install.sh
```diff
@@ -2,10 +2,19 @@
...
+  INSTALL_DIR="$XDG_DATA_HOME/fnm"
+elif [ "$OS" = "Darwin" ]; then
+  INSTALL_DIR="$HOME/Library/Application\ Support/fnm"
```
page-down, line 13: You need to remove the extra `\`, because double quotes are already used. Otherwise the path is wrong.

page-down (commented):
The path under macOS is wrong and needs to be fixed.

.ci/install.sh
jonz94: Thanks for the tips!

I did try `test -n` before, but I tried it without double quotes:

```
elif [ -n $XDG_DATA_HOME ]; then
```

and in some edge case (e.g. `XDG_DATA_HOME=""`), this will cause unexpected result.

jonz94: Now I understand that string wrapped by double quotes is very different that without double quotes in bash,
and it's important to know when to wrap string by double quotes or not.

.ci/install.sh
jonz94: Thanks for pointing out my mistake.

.ci/install.sh
ljharb: There’s no convention on Macs for ~/.local to even exist at all/

sarmong:
Will this be ever merged?

Schniz:
I'm gonna update this PR so it'll run CI

page-down:
@Schniz
The PR has been merged, however XDG_DATA_HOME is still not respected under macOS.
```text
https://github.com/Schniz/fnm/issues/622
```

This results in fnm being installed to `~/.local/share/fnm`, however the `FNM_DIR` given by `fnm env` is still `~/Library/Application Support/fnm`.
Since we already have `src/directories.rs`, it seems relatively straightforward.
Do you have any other concerns?

PR #616, opened 2021-12-29
Use on cd by default

matthieubosquet:
I noticed that there is no instructions to get fnm to autoswitch envs when navigating to a directory and tinkered a while before I found the option in the Powershell section (obviously autocomplete isn't much help either).

Could we add the `--use-on-cd` flag to all the default instructions?

I just added the flag to bash because I tested it and it worked.

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -155,12 +155,12 @@ setup_shell() {
     echo ""
     echo '  # fnm'
     echo '  export PATH='"$INSTALL_DIR"':$PATH'
-    echo '  eval "`fnm env`"'
+    echo '  eval "`fnm env --use-on-cd`"'
 
     echo '' >>$CONF_FILE
     echo '# fnm' >>$CONF_FILE
     echo 'export PATH='$INSTALL_DIR':$PATH' >>$CONF_FILE
-    echo 'eval "`fnm env`"' >>$CONF_FILE
+    echo 'eval "`fnm env --use-on-cd`"' >>$CONF_FILE
 
   elif [ "$CURRENT_SHELL" = "fish" ]; then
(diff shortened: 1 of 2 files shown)


Schniz:
Do you think we should update the installation script as well?

matthieubosquet:
@Schniz absolutely, adding it to the install script sounds good.

I just tested with zsh and the option also worked like a charm. Just added [a second commit](https://github.com/Schniz/fnm/pull/616/commits/6cc7c9b6e1cfb450dd8db2d06470433833bc2bd9) with `--use-on-cd ` by default everywhere (I just assume it also works in fish even though I haven't tried).

Schniz:
Sorry if I was misleading. It was just a genuine question, not a review of the PR.. I’m not sure if this is a good path as use-on-cd can be annoying if you don’t want it. In Bash we hijack `cd` which people may dislike, and automatic installations probably don’t need it at all. 

I wonder if more people would think this is a good idea. And if not, then how can we let more people know that this feature exists and easy to set up.

ljharb:
doing it by default is something i definitely think is a bad idea - it's far too intrusive. nvm lists mechanisms for users to do it themselves in its readme, and this seems like it's been sufficient for a long time.

matthieubosquet:
I get what you mean.

I reverted the install script and tried to make the readme a bit more explicit by mentioning the `nvm use` command and the `--use-on-cd` option. Let me know what you think.

matthieubosquet:
@Schniz did you get a chance to check out my [README edits](https://github.com/Schniz/fnm/pull/616/files)?
I think it makes it a bit more obvious what happens and what is possible...

Schniz (approved):
looks good! thank you. sorry for the delay!

Issue #575, opened 2021-11-16
fnm env substantially slower in 1.28.0


samhh:
Hi there!

I noticed that `fnm env` is a lot slower to run on my system in 1.28.0 than 1.27.0. This appears to happen irrespective of shell or terminal. I'm on Arch Linux.

`time fnm env`:

1.27.0:

```
set -gx PATH "/tmp/fnm_multishells/22708_1637100439784/bin" $PATH;
set -gx FNM_MULTISHELL_PATH "/tmp/fnm_multishells/22708_1637100439784";
set -gx FNM_DIR "/home/<user>/.fnm";
set -gx FNM_LOGLEVEL "info";
set -gx FNM_NODE_DIST_MIRROR "https://nodejs.org/dist";
set -gx FNM_ARCH "x64";

________________________________________________________
Executed in   17.98 millis    fish           external
   usr time    1.89 millis  272.00 micros    1.62 millis
   sys time    0.04 millis   36.00 micros    0.00 millis
```

1.28.0:

```
set -gx PATH "/tmp/fnm_multishells/24079_1637100477734/bin" $PATH;
set -gx FNM_MULTISHELL_PATH "/tmp/fnm_multishells/24079_1637100477734";
set -gx FNM_DIR "/home/<user>/.fnm";
set -gx FNM_LOGLEVEL "info";
set -gx FNM_NODE_DIST_MIRROR "https://nodejs.org/dist";
set -gx FNM_ARCH "x64";

________________________________________________________
Executed in  600.78 millis    fish           external
   usr time   80.04 millis  350.00 micros   79.69 millis
   sys time  300.36 millis   47.00 micros  300.32 millis
```

Schniz:
Oh! That's bad. Let's see if it relates to a recent change I made. Can you try benchmarking `fnm env --shell=fish` vs `fnm env`? I believe this is the main issue.

Schniz:
This does change a lot on my machine. STILL SLOWER THOUGH. I have to debug this.

Schniz:
It's almost 1am here, so I take the liberty to go to bed. I will release a new version tomorrow. Sorry that it happened and BIG THANK YOU for filing the issue. You're a life saver!

PR #625, opened 2022-01-05
Revert from using sysinfo, but only on Unix machines

### LINKED ISSUE #621: Shell inference slow on Linux?
Continuing the conversation on Twitter, here are the results when `shell` argument is specified and not specified for fish and bash.

For bash

```
$ time eval "$(fnm env)"

real	0m0.501s
user	0m0.004s
sys	0m0.000s

$ time eval "$(fnm env --shell=bash)"

real	0m0.002s
user	0m0.001s
sys	0m0.000s
```

For fish

```
# ~ : time fnm env --shell=fish

________________________________________________________
Executed in    1.40 millis    fish           external
   usr time    0.35 millis  350.00 micros    0.00 millis
   sys time    1.13 millis   45.00 micros    1.09 millis

# ~ : time fnm env

________________________________________________________
Executed in  517.84 millis    fish           external
   usr time    0.45 millis  454.00 micros    0.00 millis
   sys time    3.45 millis   58.00 micros    3.39 millis
```
Results are from running fnm version 1.29.1 on Endeavour/Arch Linux. System has AMD Ryzen 9 3950X and SSD.

Schniz:
Seems that sysinfo is sometimes slower than using plain `ps`.
Let's revert the change for Unix systems and benchmark afterwards.

* Fixes #621

### DIFF
--- a/src/shell/infer/mod.rs
+++ b/src/shell/infer/mod.rs
@@ -8,7 +8,7 @@ pub use self::unix::infer_shell;
 pub use self::windows::infer_shell;
 
 pub(self) fn shell_from_string(shell: &str) -> Option<Box<dyn super::Shell>> {
-    use super::*;
+    use super::{Bash, Fish, PowerShell, WindowsCmd, Zsh};
     match shell {
         "sh" | "bash" => return Some(Box::from(Bash)),
         "zsh" => return Some(Box::from(Zsh)),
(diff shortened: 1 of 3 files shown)


samhh:
Just wanted to mention that I think this PR fully resolved what I reported in #575. Thank you!

PR #626, opened 2022-01-05
Create the multishells directory if it is missing

### LINKED ISSUE #624: fnm current version reset on sleep / lock / stale terminal
Since I installed the latest fnm to try the new recursive version strategy I have the following problem:

When I lock my mac and put it into sleep for some time, my current shell sessions (including vscode) cannot find node & npx anymore. I think the symlink is getting cleaned up or something? When I open up a new terminal fnm is working properly again. But the current shell does not have any current node version. fnm current outputs none.

My setup is as simple as the following as first line in `.zshrc`:
```
eval "$(fnm env --use-on-cd --version-file-strategy recursive)"
```

Schniz:
This should fix #624, but only partially. Node versions will still be stale, but running `fnm use` again will work.
The real solution for #624 is to understand why temp directories are being deleted.

### DIFF
--- a/src/commands/use.rs
+++ b/src/commands/use.rs
@@ -93,6 +93,14 @@ impl Command for Use {
             outln!(config, Info, "{}", message);
         }
 
+        if let Some(multishells_path) = multishell_path.parent() {
+            std::fs::create_dir_all(multishells_path).map_err(|_err| {
+                Error::MultishellDirectoryCreationIssue {
+                    path: multishells_path.to_path_buf(),
+                }
+            })?;
+        }
+
         replace_symlink(&version_path, multishell_path).context(SymlinkingCreationIssue)?;
 
         Ok(())
@@ -220,6 +228,8 @@ pub enum Error {
         "Check out our documentation for more information: https://fnm.vercel.app"
     ))]
     FnmEnvWasNotSourced,
+    #[snafu(display("Can't create the multishell directory: {}", path.display()))]
+    MultishellDirectoryCreationIssue { path: std::path::PathBuf },
 }
 
 #[derive(Debug, Snafu)]


pyronaur:
@Schniz I noticed something similar happening on my  weird is happening on Macbook Pro 14" M1 Max.


I was using fnm 1.28.2 when I wrote this, and updating to  1.29.2 does allow me to do `fnm use`, so it's fixed in that sense. Just thought to leave this here anyway, in case any of that proves useful:


**On fnm 1.28.2:**


I run my node svelte compiler fine for a couple of minutes until it crashes for some reaon. When it does, it seems that node is no longer available:

```
✗ node -v
zsh: command not found: node
✗ which node
node not found
✗ which fnm
/opt/homebrew/bin/fnm
 ✗ fnm list
* v16.13.1 default
* system
✗ fnm current
none
```

When this happens, I can't make fnm use node in any way.
```
FNM_LOGLEVEL=all fnm use 16
Using Node v16.13.1
error: Can't create the symlink: No such file or directory (os error 2)
```

Just in case, tried installing another node version, same issue:

```
✗ fnm use 14
Can't find an installed Node version matching v14.x.x.
Do you want to install it? answer [y/n]: y
Installing Node v14.18.2 (x64)
Using Node v14.18.2
error: Can't create the symlink: No such file or directory (os error 2)
```

The only solution is either to open a new tab, or do this:

```
✗ eval "$(fnm env --use-on-cd)"
✗ fnm use 16
Using Node v16.13.1
```


Also tried (and failed):

* Running from Terminal instead of iTerm2
* Running iTerm2 in Rosetta mode

Schniz:
That’s odd. I wonder what makes the temp directories to be removed. It does not happen on any of the machines I had so far. Is there a setting for temp directory cleanup time?

pyronaur:
I would place my bets on either Mac 12.1 Monterey either cleaning something up, or MacOS in general to optimize for M1.

All I know I have a fresh machine, I don't have any invasive tools installed that would tamper with the tmp directory.

But if it's a system level tmp directory that fnm is using here - perhaps a fix is to just move it to a a path owned by `fnm` so that the system won't automatically optimize it? That would probably also mean owning the responsibility of then cleaning it up too, but.... 🤷  - at least that sounds simpler than figuring out what OS level magic is happening here.

Schniz:
> Mac 12.1 Monterey

Maybe that's what's going on. I will update too this evening :)

> That would probably also mean owning the responsibility of then cleaning it up too, but.... 🤷 - at least that sounds simpler than figuring out what OS level magic is happening here

This is what I had in mind too, but it is very problematic to implement. When do we clean up? Feels like `fnm env` is what should do it, but then--how do we know what to cleanup? Feels like getting the "accessed time" from the file should be enough, but it isn't, as I don't see changes when I call `stat $FNM_MULTISHELL_PATH`. The other option that makes sense is to do something with pids and check which aren't available, but this is also an issue I have on #609 where I don't know what pid to choose as I don't know what is the current shell that is being used.

Moreover, I think that tmp directories are not OS magic, they are something we should rely on... and we need to figure out what's going on 😿

pyronaur:
> Moreover, I think that tmp directories are not OS magic, they are something we should rely on

Just like cache isn't magic, as long as you keep in mind that it's temporary. Temporary by definition is......temporary - meaning if you don't know exactly when it's cleaned, assume it's always in a clean state. No idea - maybe OS temp directories have a different definition though :) 

> When do we clean up? 
When do you want the OS to clean up? Personally, I wouldn't mind running a `fnm cleanup` command, which of course would break something for someone else.

PR #637, opened 2022-01-16
[Documentation] Adds Additional info regarding .node-version

uchihamalolan:
### Why

I felt this info will enable the repo to have all the relevant information for what to be done in order to make `fnm` set the node version automatically (apart from adding `--use-on-cd` to eval statement)

For new users and users who are not familiar of `.node-version`


Schniz (approved):
That's amazing. I've been doing this all the time and never think about adding that. Thanks!

Issue #622, opened 2022-01-04
XDG_DATA_HOME is not respected under macOS


page-down:
Support for $XDG_DATA_HOME was added in a recent release, however it is not used under macOS and falls back to the default location  `~/Library/Application Support`.

fnm 1.29.1

```shell
mv ~/.fnm ~/.local/share/fnm
fnm env
# FNM_DIR -> ~/Library/Application Support/fnm
```

However $XDG_DATA_HOME is respected under Linux.

```text
# macOS
XDG_DATA_HOME=/tmp fnm env
# FNM_DIR -> ~/Library/Application Support/fnm

# Linux
XDG_DATA_HOME=/tmp fnm env
# FNM_DIR -> /tmp/fnm
```
The XDG_DATA_HOME environment variable should be used first if it is configured, and then fall back to the macOS location.

For example, the changes to the install script in Pull Request 614.

https://github.com/Schniz/fnm/pull/614

```bash
if [ -d "$HOME/.fnm" ]; then
  INSTALL_DIR="$HOME/.fnm"
elif [ ! -z $XDG_DATA_HOME ]; then
  INSTALL_DIR="$XDG_DATA_HOME/fnm"
elif [ "$OS" = "Darwin" ]; then
  INSTALL_DIR="$HOME/Library/Application Support/fnm"
  # ...
fi
```

Schniz:
I should have changed the name of the feature. It is not `XDG_DATA_HOME`, but the standard data locations for each OS. `~/Library/Application Support/fnm` seems to be the standard per https://docs.rs/dirs/latest/dirs/fn.data_dir.html

do you think this is wrong?

page-down:
In macOS environment, XDG_DATA_HOME does not exist, so there is no problem to use the path under Library as the default fnm folder.

However, if the XDG_DATA_HOME environment variable exists, my recommendation is to uniformly honor this.

Because some developers work in cross-platform environments, they may set this environment variable themselves and potentially use unified dotfiles to ensure consistency across environments.

I hope you understand this usage requirement and reduce fragmentation of fnm behavior, thanks. I would like to reopen this issue if you think it is worth it.

Schniz:
> It seems that I cannot reopen the issue on my own.

didn't know that. sorry.

I understand your frustration, but I also agree with https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-478337412

[...]

PR #638, opened 2022-01-16
Use local cache directory instead of temp directory for symlinks

### LINKED ISSUE #624: fnm current version reset on sleep / lock / stale terminal

Schniz:
still with no cleanup, probably need to have --cleanup-duration=... for fnm env
or another command for cleaning up old symlinks.

Fixes #624

### DIFF
--- a/src/commands/env.rs
+++ b/src/commands/env.rs
@@ -6,6 +6,7 @@ use crate::path_ext::PathExt;
 use crate::shell::{infer_shell, Shell, AVAILABLE_SHELLS};
 use colored::Colorize;
 use std::fmt::Debug;
+use std::path::PathBuf;
 use thiserror::Error;
 
 #[derive(clap::Parser, Debug, Default)]
@@ -30,10 +31,16 @@ fn generate_symlink_path() -> String {
     )
 }
 
+fn symlink_base_dir() -> PathBuf {
+    match dirs::cache_dir() {
+        Some(cache_dir) => cache_dir.join("fnm").join("multishells"),
+        None => std::env::temp_dir().join("fnm_multishells"),
+    }
+    .ensure_exists_silently()
+}
+
 fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
-    let base_dir = std::env::temp_dir()
-        .join("fnm_multishells")
-        .ensure_exists_silently();
+    let base_dir = symlink_base_dir();
     let mut temp_dir = base_dir.join(generate_symlink_path());
 
     while temp_dir.exists() {
--- a/src/commands/env.rs
+++ b/src/commands/env.rs
@@ -32,11 +32,7 @@ fn generate_symlink_path() -> String {
 }
 
 fn symlink_base_dir() -> PathBuf {
-    match dirs::cache_dir() {
-        Some(cache_dir) => cache_dir.join("fnm").join("multishells"),
-        None => std::env::temp_dir().join("fnm_multishells"),
-    }
-    .ensure_exists_silently()
+    crate::directories::multishell_storage().ensure_exists_silently()
 }
 
 fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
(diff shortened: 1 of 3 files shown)


Schniz:
btw I've been using this patch since yesterday and it fixes stuff for me, that's why I want to release it 😁

alexeyten:
`$XDG_RUNTIME_HOME` could be too restrictive for someone.
> [...] The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.
> The lifetime of the directory MUST be bound to the user being logged in.

May be `$XDG_CACHE_HOME` is good enough solution.

page-down:
> ... XDG_CACHE_HOME ... tmp directories ...
> I wonder if I should change it just for MacOS(>12.1)  ... or make it consistent

In my opinion, the cache/tmp places are not very suitable, because the files in these places mean that they can be cleaned up, however we have these symbolic links that need to be maintained.
So it should take effect on all systems. I think this is a meaningful improvement.

> I can go with dirs::runtime_dir().or_else(dirs::cache_dir) ...
> ... and dirs::runtime_dir does not have a MacOS default ...

Maybe you have no problem with the cache directory now, but you don't want to do it all over again if a future user reports that the files placed in the cache has been purged.

IMHO, The dir location recommended by alex (`$XDG_STATE_HOME or $XDG_RUNTIME_DIR`) is a good fit.

My recommended priorities for folders are:
$XDG_STATE_HOME -> dirs::state_dir() -> $XDG_RUNTIME_DIR -> dirs:: runtime_dir()
-> $XDG_DATA_HOME -> dirs:: data_dir() -> $XDG_CACHE_HOME -> dirs::cache_dir()

As I mentioned in the following issue, please try to respect environment variables on all platforms. If a user sets an XDG environment variable that doesn't exist by default on macOS, then he expects this to be used. Using consistent behavior across platforms is also good for project maintenance.

https://github.com/Schniz/fnm/issues/622

Maybe you don't have time yet and the issue above is still waiting for your response, please consider reopening it as well.

In my opinion crate `dirs` focus on providing standard directory paths, but that doesn't mean that software developers don't need to pay attention to environment variables.

uloco:
FWIW the /tmp directory in mac only cleans up files that have not been touched since 3 days (per default). I don't know if this also relates to symbolic links and how this works. But generally I think it would be a good idea to clean up the unused symlinks from time to time. My local cache (workaround to ~/.tmp) is getting bigger and bigger like this. The script that is cleaning up the /tmp directory on MacOS lies under `etc/periodic/{daily,weekly,monthly}` and the config is under `etc/defaults/periodic.conf`

page-down:
> But generally I think it would be a good idea to clean up the unused symlinks from time to time.

Yes, this needs to be considered and handled as soon as the symbolic links are moved out of the default $TMPDIR.

Is there a cleanup of these files in other platforms now?

Schniz:
> FWIW the /tmp directory in mac only cleans up files that have not been touched since 3 days (per default).

I feel like it's a regression in 12.1. This is why my commit mentions:

> still with no cleanup, probably need to have --cleanup-duration=... for fnm env
> or another command for cleaning up old symlinks.

I really dislike that it becomes something we need to consider ourselves, would rather use the platform 😭

uloco:
There is a difference between $TMPDIR and /tmp. Your symlinks in mac are in /var/folders/... (where $TMPDIR) points to. $TMPDIR is not cleaned by the periodic script, but /tmp is.

dragonfire1119:
I've been dealing with this bug since Monterrey do you think this will be merged soon? Is there a temporary workaround?

Schniz:
I hope to merge it by tomorrow or Sunday. Sorry about that.

dragonfire1119:
> I hope to merge it by tomorrow or Sunday. Sorry about that.

Cool thanks for your time and making fnm.

Issue #441, opened 2021-04-25
fnm env: Use fish_add_path in fish shell


jonahsnider:
`fnm env` should use the [`fish_add_path`](https://fishshell.com/docs/current/cmds/fish_add_path.html) function when appropriate. This avoids appending multiple fnm binary paths to `$PATH`, which can happen in a child shell.

```diff
diff --git a/a.fish b/b.fish
index 3a9ba79..c299b09 100644
--- a/a.fish
+++ b/b.fish
@@ -1,4 +1,4 @@
-set -gx PATH "/tmp/fnm_multishell_16129_1619310866524/bin" $PATH;
+fish_add_path -g "/tmp/fnm_multishell_16129_1619310866524/bin";
 set -gx FNM_MULTISHELL_PATH "/tmp/fnm_multishell_16129_1619310866524";
 set -gx FNM_DIR "/home/username/.fnm";
 set -gx FNM_LOGLEVEL "info";
```

Schniz:
Interesting. What is the difference? I guess we should have some grace period. The current solution does not work on latest fish?

jonahsnider:
The current solution works fine by modifying `PATH` which fish exposes for compatibility. The only advantage `fish_add_path` offers is to avoid duplicate entries in the `PATH` as well as modifying the `fish_user_paths` (which is automatically prepended to `PATH`).

This PR would just make the fish script more like an idiomatic fish script.

rootblack45:
`fish_add_path` did not help to avoid the script to append a new fnm binary path in this situation. Because fnm will create a new binary path whenever you run the `fnm env` command. I think the best way is we need to check the `PATH` variable before it calls set function or erases it from `PATH` before adding a new one.

Schniz:
I'm working on a feature that will make less mess. Then, we can use `fish_add_path` and it will be the same path and not a random one.
Still, I wonder what's the adoption process for fish. Should we have some kind of `if fish_version > 3.2.0 then ... else ...`?

rootblack45:
I think we can do it simply by using `contains` like the example from [fish 2.0](https://fishshell.com/docs/2.0/commands.html#contains). It's gonna be like this:

```bash
if not contains $FNM_MULTISHELL_PATH/bin $PATH
    set -gx PATH "$FNM_MULTISHELL_PATH/bin" $PATH
end
```

Schniz:
PRs welcome 😉

PR #641, opened 2022-01-18
Fix duplicate variable of `multishell_path` in `PATH`

0x221A:
This PR contains the change of the `env` command and `shell`.

See: #441


Schniz:
Hi! thanks for the contribution.

I'm all in for filtering the PATH to not add directories that are not necessary and just messes up everything. Do you think it should be a shell code or can it be in a single place? We can get `PATH` with `std::env` and solve it once for all shells

> This change will avoid the `env` command generating new `multishell_path` whenever it was exist in the fnm config

I don't understand why doing this: is there a reason of calling `fnm env` if you don't want to set up a new shell?
Can you please explain the use case? 😃 

Personally I use tmux in zsh, so when my `.zshrc` file initializes on the main shell, it creates a FNM_MULTISHELL_PATH for the global shell, and then it creates a new multishell_path for each tmux pane/tab/window I open. So with this code, I will have to conditionally evaluate fnm if I'm within tmux, or to do something like `eval "$(FNM_MULTISHELL_PATH= fnm env)"`

rootblack45:
> I'm all in for filtering the PATH to not add directories that are not necessary and just messes up everything. Do you think it should be a shell code or can it be in a single place? We can get PATH with std::env and solve it once for all shells

I think it can be in a single place in `env.rs`. The code will be like this:

```rust
fn dedup_and_merge_path_variable(path: &Path) -> String {
    let current_path = std::env::var_os("PATH").expect("Can't read PATH env var");
    let mut split_paths: Vec<_> = std::env::split_paths(&current_path).collect();
    split_paths.dedup_by(|a, b| a == b || a.to_str().unwrap().contains("fnm_multishells"));
    split_paths.insert(0, path.to_path_buf());
    let new_path = std::env::join_paths(split_paths).expect("Can't join paths");
    new_path.to_str().expect("Can't read PATH").to_owned()
}

impl Command for Env {
    ...
    fn apply(self, config: &FnmConfig) -> Result<(), Self::Error> {
        ...
        println!(
            "{}",
            shell.set_env_var("PATH", dedup_and_merge_path_variable(&binary_path).as_str())
        );
        ...
    }
}
```

> I don't understand why doing this: is there a reason of calling fnm env if you don't want to set up a new shell?
Can you please explain the use case?

My point is whenever we need to reload the session the shell setup command will run. From the previous codes when we invoke the `make_symlink` function it returns a path of the symlink and uses that path to set into `PATH` and `FNM_MULTISHELL_PATH`. Hence if `FNM_MULTISHELL_PATH` was already in the config we don't need to make a new symlink. This will be an easier way to handle the duplicate `PATH` variables and we can guarantee that the path variable in `PATH` and `FNM_MULTISHELL_PATH` are the same.

You can try by opening a terminal and trying to run `source .zshrc` multiple times you will see that from the previous code `FNM_MULTISHELL_PATH` will change every time.

Schniz:
> My point is whenever we need to reload the session the shell setup command will run.

how many times do you find yourself reloading a session?
(by reloading I assume you mean `source ~/.zshrc`)

src/commands/env.rs
```diff
@@ -44,6 +45,15 @@ fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
...
 }
 
+fn dedup_and_merge_path_variable(path: &Path) -> String {
+    let current_path = std::env::var_os("PATH").expect("Can't read PATH env var");
+    let mut split_paths: Vec<_> = std::env::split_paths(&current_path).collect();
+    split_paths.dedup_by(|a, b| a == b || a.to_str().unwrap().contains("fnm_multishells"));
+    split_paths.insert(0, path.to_path_buf());
+    let new_path = std::env::join_paths(split_paths).expect("Can't join paths");
+    new_path.to_str().expect("Can't read PATH").to_owned()
+}
```
Schniz, line 55: can we avoid `expect` by returning a `Result`? We are using `thiserror` so we can have nice error reports instead of panics

src/commands/env.rs
```diff
@@ -44,6 +45,15 @@ fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
     temp_dir
 }
 
+fn dedup_and_merge_path_variable(path: &Path) -> String {
+    let current_path = std::env::var_os("PATH").expect("Can't read PATH env var");
+    let mut split_paths: Vec<_> = std::env::split_paths(&current_path).collect();
+    split_paths.dedup_by(|a, b| a == b || a.to_str().unwrap().contains("fnm_multishells"));
+    split_paths.insert(0, path.to_path_buf());
```
Schniz, line 52: if we would collect into a `std::collections::Set`, and then add the pathbuf, we would get dedup for "free", right?
or if it is harder to collect into a set, we can insert and then simply dedup, instead of dedup and then insert?

src/commands/env.rs
```diff
@@ -62,13 +72,23 @@ impl Command for Env {
...
             .ok_or(Error::CantInferShell)?;
-        let multishell_path = make_symlink(config);
+
+        let multishell_path = {
+            if config.multishell_path().is_none() {
+                make_symlink(config)
+            } else {
+                config.multishell_path().unwrap().to_path_buf()
+            }
+        };
```
Schniz, line 82: as I mentioned in a previous comment, I am not sure about it. Maybe this could be a flag (`--force-create-symlink` or `--ignore-current-env`, not sure about the "polarity" or the namings tho)

anyway, a more rusty way of doing that, would be:

```suggestion
        let multishell_path = match config.multishelll_path() {
            None => make_symlink(config),
            Some(p) => p.to_path_buf(),
        };
```

or

```suggestion
        let multishell_path = if let Some(p) = config.multishell_path() {
            p.to_path_buf()
        } else {
            make_symlink(config)
        };
```

or something with `.map` or `.unwrap_or_else` and all these `Option` functions 😁

Schniz: what do you think about introducing a flag? what polarity do you think it should have?
I lean towards `--ignore...` because it's an optimization: if we have `--force-...` then all tmux users (or any other terminal multiplexer) will share the node version across all shell sessions

Schniz:
I don't say that enough, but I'm really thankful for the contribution. I hope my comments and questions don't annoy you 🙏

src/commands/env.rs
rootblack45: I have added 2 more error reports. The error message is based on the previous code. But one more thing is should I handle `to_str() or just use `unwarp()`? What do you think?
```rust
pub enum Error {
    ...
    #[error("Can't read PATH env var")]
    CantReadPath,
    #[error("Can't join paths: {}", source)]
    CantJoinPaths { source: std::env::JoinPathsError },
}
```

src/commands/env.rs
rootblack45: Ok, Here is the code that uses `std::collections::HashSet`. What do you think? 
```rust
let mut split_paths: HashSet<_> = std::env::split_paths(&current_path)
        .filter(|p| !p.to_str().unwrap().contains("fnm_multishells"))
        .collect();
split_paths.insert(path.to_path_buf());
```
I want to dedup it first to remove the `fnm_multishells` path that doesn't same as the value from `FNM_MULTISHELL_PATH`

src/commands/env.rs
rootblack45: I think I'm too worried about the temp dir 😅 So I will revert the code and focus on the duplicate path variable.

rootblack45:
@Schniz No problem 😁. This is my first time contributing to the Rust project. Thanks for your advice.

src/commands/env.rs
```diff
@@ -44,6 +46,17 @@ fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
     temp_dir
 }
 
+fn merge_path(path: &Path) -> Result<String, Error> {
+    let current_path = std::env::var_os("PATH").ok_or(Error::CantReadPath)?;
+    let mut split_paths: HashSet<_> = std::env::split_paths(&current_path)
+        .filter(|p| !p.to_str().unwrap().contains("fnm_multishells"))
```
Schniz, line 52: 👀 what if the name would not be `fnm_multishells`? or someone will have a `/home/user/fnm_multishells_are_annoying/my-bin`? 😜

Maybe we should filter out the input `FNM_MULTISHELL_PATH` and then we can call it a day? https://github.com/Schniz/fnm/blob/74f60b3590ce93d63e28349cb4ca0116cf23702c/src/config.rs#L33

I'm thinking here with you, I don't have the answers hehe 😃 
again, sorry if that's annoying.

src/commands/env.rs
```diff
@@ -44,6 +46,17 @@ fn make_symlink(config: &FnmConfig) -> std::path::PathBuf {
...
+        .collect();
+    split_paths.insert(path.to_path_buf());
+    let new_path =
+        std::env::join_paths(split_paths).map_err(|source| Error::CantJoinPaths { source })?;
```
Schniz, line 56: I hope that this works 🤔  what will happen if a Windows user runs Bash in cygwin or other emulators, and eventually call the Windows `.exe` instead of the Linux binary? 🤔 
If we need to join it ourselves by having a function in the trait object, it is going to be a bummer 🐑

src/commands/env.rs
```diff
@@ -62,13 +75,15 @@ impl Command for Env {
...
+
         let multishell_path = make_symlink(config);
         let binary_path = if cfg!(windows) {
             multishell_path.clone()
         } else {
             multishell_path.join("bin")
         };
-        println!("{}", shell.path(&binary_path));
+        let merged_path = merge_path(&binary_path)?;
+        println!("{}", shell.set_env_var("PATH", merged_path.as_str()));
```
Schniz, line 86: I think that not all shells are using the same env var expression. But I'm not sure. I don't really know why we had `shell.path(..)` instead of `set_env_var` 😁

Schniz (commented):
hey there! thanks for all the hard work. please check my comment (especially about the `fnm_multishells` filtering)

thanks again!

src/commands/env.rs
rootblack45: Does it possible? Because every time `env` was invoked the temp dir will be created based on `$TMPDIR/fnm_multishells`
https://github.com/Schniz/fnm/blob/52a9cfedd1327b9d7a828d8ecc1b84077ca8c6fa/src/commands/env.rs#L35-L37
IMO Filter it out by using the value from `std::env::temp_dir().join("fnm_multishells")` is better

src/commands/env.rs
rootblack45: It's hard to say that it works. It has a lot of problems if the Windows user runs `bash` in Cygwin or other emulators. 

The first problem I just found is the value of the `PATH` variable. The value of the `PATH` variable from Cygwin looks different from the output from fnm. I think the only way to solve this problem is to revert the code to 33c8da27f84c074fc9bc4836d4985b613fbbb269. And join `PATH` by ourselves for every cross-platform shell.

Another problem is the result from `make_symlink()` is based on what OS is using. In the current version of fnm, it still generated Windows path style even if we're using bash. I think this should be a new issue.

src/commands/env.rs
rootblack45: i'm not sure either 😁

src/commands/env.rs
Schniz: Well we can do that for now. When I land the location change I will have to change this line to conform the new location too but that's cool.

src/commands/env.rs
Schniz: Seems that tests fail. Also, seems like Fish has a different way of adding a PATH entry

src/commands/env.rs
rootblack45: @Schniz What do u think about the above comment. In addition, I would recommend for the users that use PowerShell in WSL2 to use Linux Binary of `fnm` instead of Windows `.exe`

src/commands/env.rs
rootblack45: @Schniz Fixed

src/commands/env.rs
rootblack45: Should we add this recommendation to the document?

Schniz: I think we should 👀

Schniz: good catch 🙌

rootblack45:
I'm not sure why the `feature_tests::aliases::PowerShell` test was passed and sometimes it was failed

rootblack45:
@Schniz Hi, according to this commit https://github.com/Schniz/fnm/commit/ad0ecd7fff864430e072c0df2fe984ef0875324b. I just updated the code to support the local cache directory.

rootblack45:
@Schniz I just updated the code according to this commit https://github.com/Schniz/fnm/commit/c85ff97f1404b0042f7378acb0ad67b3fd607189

Schniz:
🙏 
let's see if tests pass 🙏

rootblack45:
> 🙏 let's see if tests pass 🙏

It's failed. Look like I need to install Windows to test it...

Schniz:
I wish GitHub Codespaces had a Windows option. 😢

rootblack45:
@Schniz Just fixed. I hope this is the last 😂

rootblack45:
Finally, it passed 🎉

src/shell/bash.rs
```diff
@@ -16,7 +16,23 @@ impl Shell for Bash {
...
+                new_path="{path}"
+                IFS=":"
+                for p in $PATH; do
+                    if ! echo $p | grep -q {multishell_storage}; then
+                        new_path="$new_path:$p"
+                    fi
+                done
+                unset IFS
+                export PATH=$new_path
+                unset new_path
```
Schniz, line 31: hmmmm. what if a user had a `IFS` variable prior to `eval "$(fnm env)"`? it would probably be unset, right? (I'm not sure, but I assume it shares the same context)

Can we make these parts written in Rust instead of bash/other shell scripting language?
I want to have the minimum amount of shell code in fnm

Schniz (commented):
I'm really sorry that I'm commenting _after_ the PR is passing tests. bah. I really appreciate your contribution but I am not sold on having a big chunk of bash/zsh/fish/etc code for setting the `PATH`

what do you think?
can we make it Rusty and keep all initializations as `export PATH=${THE NEW PATH}`? Because I think that the complexity is very high right now.

src/shell/bash.rs
rootblack45: Yes, we can, but we have to manage `split_paths()` and `join_paths()` manually. Due to different types of path separators between OS.

> hmmmm. what if a user had a IFS variable prior to eval "$(fnm env)"? it would probably be unset, right? (I'm not sure, but I assume it shares the same context)

`IFS` already has a default value. However, after we unset it, Bash will process word extraction using its default value.

PR #667, opened 2022-02-12
Make fnm create global scoped environment variables in PowerShell

### LINKED ISSUE #664: PowerShell bug & enhancement
Bug: On Windows PowerShell 5.1 (Default version on Windows 10/11), after setting up with `fnm env --use-on-cd | Out-String | Invoke-Expression`, PowerShell will complain with the following error message when using `cd` without any argument
```
Set-Location : Cannot process argument because the value of argument "path" is null. Change the value of argument
"path" to a non-null value.
```
Enhancement: I would like to set up fnm manually instead of automatically every time I open a new session.  I tried adding this function to my profile and calling it on demand. But this fails because fnm does not explicitly operate in global scope
```PowerShell
function usefnm {
  fnm env --use-on-cd | Out-String | Invoke-Expression
}
```

Proposed solution to both: change the following code https://github.com/Schniz/fnm/blob/c85ff97f1404b0042f7378acb0ad67b3fd607189/src/shell/powershell.rs#L39-L44
to
```PowerShell
function global:Set-FnmOnLoad {{ {autoload_hook} }} 
function global:Set-LocationWithFnm {{ param($path); if ($path -eq $null) {{Set-Location}} else {{Set-Location $path}}; Set-FnmOnLoad }}
Set-Alias -Scope global cd_with_fnm Set-LocationWithFnm
Set-Alias -Option AllScope -Scope global cd Set-LocationWithFnm
Set-FnmOnLoad
```
This works on both PowerShell 5.1 and 7.2

citreae535:
Close #664

### DIFF
--- a/src/shell/infer/windows.rs
+++ b/src/shell/infer/windows.rs
@@ -1,7 +1,6 @@
 #![cfg(not(unix))]
 
 use crate::shell::Shell;
-use log::debug;
 use std::ffi::OsStr;
 use sysinfo::{ProcessExt, System, SystemExt};
 
--- a/src/shell/powershell.rs
+++ b/src/shell/powershell.rs
@@ -36,11 +36,10 @@ impl Shell for PowerShell {
         };
         Ok(formatdoc!(
             r#"
-                function Set-FnmOnLoad {{ {autoload_hook} }}
-                function Set-LocationWithFnm {{ param($path); Set-Location $path; Set-FnmOnLoad }}
-                Set-Alias cd_with_fnm Set-LocationWithFnm -Force
-                Remove-Item alias:\cd
-                New-Alias cd Set-LocationWithFnm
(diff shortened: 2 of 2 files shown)


citreae535:
PowerShell-related tests passed but `cargo test` failed with
```
failures:
    feature_tests::current::Bash
    feature_tests::multishell::Bash
    feature_tests::uninstall::Bash
    feature_tests::use_alias_not_installed::Bash

test result: FAILED. 96 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 779.38s
```
I have fnm installed from cargo on both Windows and WSL environment. Running the tests messed up node installations in WSL. I think the tests do not work well with WSL when there's an existing installation of fnm

JohyC:
create global scoped environment variables in PowerShell is awesome idea! 
I need it...

Schniz:
I don't understand this but tests pass, so lgtm!

PR #678, opened 2022-02-23
node-build compilation fallback

Schniz:
Add the ability to add `--force-arch` to `fnm env` so if a prebuilt version is not found, fnm will use `node-build` to compile the Node.js version locally.

### DIFF
--- a/src/compile_node.rs
+++ b/src/compile_node.rs
@@ -40,7 +40,7 @@ pub fn compile_node_with_node_build(config: &FnmConfig, version: &Version) -> Re
 
     let portal_path_fixed = installation_path.to_str().unwrap().replace(' ', "\\ ");
     let version_str = version.v_str();
-    let version_str = version_str.strip_prefix("v").unwrap();
+    let version_str = version_str.strip_prefix('v').unwrap();
 
     log::debug!(
         "Going to run `node-build` to compile node {} in {}",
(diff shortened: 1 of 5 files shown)


src/compile_node.rs
```diff
@@ -0,0 +1,114 @@
...
+    },
+}
+
+pub fn compile_node_with_node_build(config: &FnmConfig, version: &Version) -> Result<(), Error> {
+    let path = {
+        let mut path = version.installation_path(config);
+        path.pop();
+        path
+    };
+    let temp_installations_dir = tempfile::tempdir().unwrap();
```
Schniz, line 37: node-build does not support spaces in directories too. We need to find a solution for that 🥲 
related: 
* https://github.com/Schniz/fnm/issues/622#issuecomment-1029320579

Schniz: I mean, the solution is to have a `dirs` module (which we already have) and probably store stuff in `.local/fnm` or something unless XDG_ env vars are provided, but I want to make the migration fast 🥲

trm217:
@Schniz Could you need help working on this? I am currently having a bit of an issue with other team members using nvm and arm64 NodeV14, I would certainly look forward not having to switch back to nvm 😂

Schniz:
I think it’s fairly finished. I’ll take another look and will release it on the next version. The only problem is node-build not supporting spaces in the installation path

Issue #390, opened 2021-02-09
Git Bash issues with CLI tools


alumni:
The setup for Git Bash (Windows) cannot run cli tools that are installed using `npm install -g`.
```
$ fnm env
export PATH="C:\\Users\\XXX\\AppData\\Local\\Temp\\fnm_multishell_15584_1612875966568":$PATH
export FNM_MULTISHELL_PATH="C:\\Users\\XXX\\AppData\\Local\\Temp\\fnm_multishell_15584_1612875966568"
...
```
Running node works:
```
node -v
v15.8.0
```
But bash is somehow confused about the actual path:
```
$ which node
/c/Users/XXX/Work/neo-backend/\Users\XXX\AppData\Local\Temp\fnm_multishell_17812_1612875949353/node
```
And if I use any cli tool installed with npm, I'll get this:
```
$ npm install -g yarn
added 1 package

$ yarn 
node:internal/modules/cjs/loader:928
  throw err;
  ^

Error: Cannot find module 'C:\Program Files\Git\Users\XXX\AppData\Local\Temp\fnm_multishell_17812_1612875949353\node_modules\yarn\bin\yarn.js'
```

This can be fixed by making sure that the `PATH` env variable is using Git-Bash paths:
```
export PATH="/c/Users/XXX/AppData/Local/Temp/fnm_multishell_15584_1612875966568":$PATH
```

alumni:
One more Git Bash issue, after setting the correct `PATH` variable:

If running `fnm use system`, `FNM_MULTISHELL_PATH`  will symlink `Z:\_fnm_\Nothing\Should\Be\Here\installation`, which does not exist. This apparently confuses Git Bash, which stops checking the other folders in the path for the actual node executable.

alumni:
I've been playing with `fnm` today and I must say it's quite fast, I like it a lot ;) 

I made a workaround for this invalid `PATH` issue. Here's my startup (`.bashrc` etc) script:

```sh
eval $(fnm env | sed 1d)
export PATH=$(cygpath $FNM_MULTISHELL_PATH):$PATH

if [[ -f .node-version || -f .nvmrc ]]; then
    fnm use
fi
```

Schniz:
what is `cygpath` 😨 can we fix it upstream?

Schniz:
[...]

PR #724, opened 2022-04-21
fix(env): use unix paths in windows bash, zsh, fish

dheerajpv:
Related to #390.

Fix environment variable generation for `fnm env` in Windows.

Cygwin, MSYS, and Git Bash use UNIX-style paths (like `/c/Users/username/AppData/Local/fnm_multishells`), but fnm generates Windows-style paths (like `C:\Users\username\AppData\Local\fnm_multishells`) in Windows, regardless of the shell.

With these changes, fnm can convert paths to UNIX using `cygpath` in bash, zsh, and fish, but leave Windows paths alone in pwsh and cmd.


src/commands/env.rs
```diff
@@ -7,7 +7,9 @@ use crate::path_ext::PathExt;
 use crate::shell::{infer_shell, Shell, AVAILABLE_SHELLS};
 use colored::Colorize;
 use std::fmt::Debug;
+use std::process::Command as SystemCommand;
```
Schniz, line 10: why do we need to import `Command`? can we use `use std::process` and `process::Command`?

src/commands/env.rs
```diff
@@ -72,7 +74,7 @@ impl Command for Env {
         println!("{}", shell.path(&binary_path)?);
         println!(
             "{}",
-            shell.set_env_var("FNM_MULTISHELL_PATH", multishell_path.to_str().unwrap())
+            shell.set_env_var("FNM_MULTISHELL_PATH", &format_path(multishell_path.to_str().unwrap(), &shell.to_string()))
```
Schniz, line 77: Can we avoid stringifying the shells?
We can add a `shell.format_path()` if we want to 👍

src/pathutils.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(windows)]
+pub fn format_path(path: &str, shell: &str) -> String {
+    match shell {
+        "cmd" | "pwsh" => String::from(path),
+        _ => {
+            let output = Command::new("cygpath")
+                .arg(path)
+                .output()
+                .expect("failed to convert Windows path to UNIX");
```
Schniz, line 12: that's cool!!!!

1. can we make the error message a little clearer so error reports will be easier for me? `Failed to convert Windows path to Unix: (actual error message)` -- most users would probably don't know that it's `cygpath` which generates the paths
2. can we make this function `std::io::Result<String>` and handle the `Err` branch in the callsite?

src/pathutils.rs
```diff
@@ -0,0 +1,25 @@
+use crate::shell::Shell;
+use std::process::Command;
+
+#[cfg(windows)]
+pub fn format_path(path: &str, shell: &str) -> String {
+    match shell {
+        "cmd" | "pwsh" => String::from(path),
+        _ => {
+            let output = Command::new("cygpath")
```
Schniz, line 9: Will this solve WSL issues too like #758? Is `cygpath` a part of WSL?

src/shell/bash.rs
```diff
@@ -51,4 +52,8 @@ impl Shell for Bash {
             autoload_hook = autoload_hook
         ))
     }
+
+    fn to_string(&self) -> String {
+        String::from("bash")
+    }
```
Schniz, line 58: I mentioned that we should probably not stringify them as it hurts the soundness of the functions (allows any kind of string). But if we can't do that, this function should probably be as:

```suggestion
    fn as_str(&self) -> &'static str {
        "bash"
    }
```

As there is no need for the extra allocation, we know that statically and we _want_ it to be static. Then we can implement `ToString` that clones the `self.as_str()` if necessary, but I don't believe it is.

src/commands/env.rs
Schniz: The fact that every shell is a different struct is weird, I will fix that in the future. This should be 100% enum.

src/shell/zsh.rs
```diff
@@ -3,8 +3,9 @@ use crate::version_file_strategy::VersionFileStrategy;
 use super::shell::Shell;
 use indoc::{formatdoc, indoc};
 use std::path::Path;
+use crate::pathutils::format_path;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
Schniz, line 8: do we need this `Clone`?

src/pathutils.rs
dheerajpv: No, `cygpath` comes from Cygwin. It's included in Cygwin, MSYS2, and Git Bash (which are the most common ways of using bash/zsh/fish on Windows natively. It doesn't exist on WSL, but from a bit of searching I found that WSL does include a `wslpath` command that functions in the same way (but adding a `/mnt` in front to make it WSL compatible) which we could probably use if we can detect if a Windows executable is being used in WSL.

src/pathutils.rs
dheerajpv: I can add anyhow context to display the actual error message.

For the second point, how would the error even be handled? I don't really see a way to proceed if the path can't be converted properly.

Schniz: We can’t continue but we don’t want to panic: we can rely on error handling

Schniz:
I forgot to say thanks in advance for this contribution. Really good stuff!

dheerajpv:
Adding new flags `--wsl` and `--cygwin` for `fnm env` might help solve #758 as well, because I can't find a great way to detect if the shell is run in something cygwin-based or WSL reliably.

Running `./target/debug/fnm.exe env --shell bash` in WSL works when launched from Windows Terminal, but if I open WSL in Git Bash (presumably works the same in other cygwin-based shells) `cygpath` is being found so it generates paths that won't work in WSL.

src/shell/powershell.rs
```diff
@@ -48,4 +49,8 @@ impl Shell for PowerShell {
     fn to_clap_shell(&self) -> clap_complete::Shell {
         clap_complete::Shell::PowerShell
     }
+
+    fn format_path(&self, path: &str) -> Result<String> {
+        Ok(path.to_string())
```
Schniz, line 54: powershell can be run on Mac and Linux (not sure about usage though). Maybe we need to call `unixpath` here too? We probably can't know if we're in WSL or Cygwin, or do we? based on the infer_shell infra and the process-walking method?
What I'm thinking is--maybe it's not shell-related, but environment related, and therefore if we have this information?

We can start without any inference, so it will require the following, if one wants to use the Windows version of fnm within Cygwin or WSL:

```
fnm env --path-formatter=cygwin
fnm env --path-formatter=wsl
```

what do you think about it?

src/shell/zsh.rs
```diff
@@ -51,4 +53,8 @@ impl Shell for Zsh {
             autoload_hook = autoload_hook
         ))
     }
+
+    fn format_path(&self, path: &str) -> Result<String> {
+        to_unix_path(path).context("Failed to convert Windows path to Unix")
```
Schniz, line 58: I don't think we need to add that context, as it's implied from the function itself

Schniz (commented):
maybe it's not a shell concern but environment one. I'm sorry that I'm requesting changes like a pendulum but I'm thinking about the problem as I'm reviewing the code. Really appreciate the contribution!

src/shell/powershell.rs
dheerajpv: Actually, I figured out a way to detect if the executable is being run from WSL, but inferring the shell is not possible and the user needs to provide it as a flag (maybe there can be a custom error message for it).
Detecting WSL works the same way as shell inferring, at some point up the parent process chain there is a process named "wsl.exe", but there's no information on the shell being used.

Schniz:
Superseded by #960 , sorry <picture data-single-emoji=":pray_parrot:" title=":pray_parrot:">:pray_parrot:</picture>

Issue #463, opened 2021-06-01
Nushell support


Tarnadas:
Now that nushell has an eval-like mechanism for setting environment variables, I’d like to open this feature request to also add support for it.

Via the `load-env` command, one can set multiple environment variables, see here:
https://github.com/nushell/nushell/issues/3481

Is that enough to make it possible or is something missing?

Schniz:
It's enough for basic usage. But if we want `--use-on-cd` we might want to add functions or whatever it is called in nushell. I really want to try nushell again so it might be a nice experiment 😃

orta:
You can get the path edit via:

```sh
/home/orta> fnm env | lines | split column '"' | first | get Column2
/tmp/fnm_multishells/10174_1623514914978/bin
```

Then the rest of the env vars via:

```
fnm env | str  find-replace -a "export " '' | str find-replace -a '"' '' |  lines | split column = | rename name value | last 5
───┬──────────────────────┬────────────────────────────────────────────
 # │         name         │                   value                    
───┼──────────────────────┼────────────────────────────────────────────
 0 │ FNM_MULTISHELL_PATH  │ "/tmp/fnm_multishells/10210_1623514981704" 
 1 │ FNM_DIR              │ "/home/orta/.fnm"                          
 2 │ FNM_LOGLEVEL         │ "info"                                     
 3 │ FNM_NODE_DIST_MIRROR │ "https://nodejs.org/dist"                  
 4 │ FNM_ARCH             │ "x64"                                      
───┴──────────────────────┴────────────────────────────────────────────
```

So to mix it all together:

```sh
#Update paths
echo $nu.path | append (fnm env | lines | split column '"' | first | get Column2) | config set_into  path

# Add env vars
fnm env | str  find-replace -a "export " '' | str find-replace -a '"' '' |  lines | split column = | rename name value | last 5 | load-env
```

And at least from some current messing around it seems to work

iamprady:
Nushell added pathvar command for dynamic path updates. To make work fnm in nushell, add these two commands in the startup:
`fnm env --shell bash | lines | last 5 | str find-replace 'export ' '' | split column = | rename name value | load-env`
`pathvar add $nu.env.FNM_MULTISHELL_PATH`
Info: Requires nushell v0.34+ (pathvar command is supported only with v0.34 and above)

fellnerse:
I had to change the path like this:
`pathvar add ($nu.env.FNM_MULTISHELL_PATH + /bin)`

[...]

PR #800, opened 2022-08-17
Added json option to env command

zaucy:
Main motivation for this PR is to make the nushell experience better (#463). Taken from one of the [suggestions in the comments](https://github.com/Schniz/fnm/issues/463#issuecomment-1132510568) I added a `--json` option.

```nushell
fnm env --json | from json | load-env
let-env PATH = ($env.PATH | prepend ($env.FNM_MULTISHELL_PATH + "/bin"))
```


src/commands/env.rs
```diff
@@ -15,6 +15,9 @@ pub struct Env {
     #[clap(long)]
     #[clap(possible_values = AVAILABLE_SHELLS)]
     shell: Option<Box<dyn Shell>>,
+    /// Print JSON instead of shell commands.
+    #[clap(long)]
+    json: bool,
```
Schniz, line 20: We should probably make it broader, so `--output=json` is preferable. `--output=json` conflicts with `--shell=ANY` which can be declared with `#[clap(conflicts_with = "shell")]`

That being said, `--shell=...` is practically `--output=...`, right? so maybe we can leverage `--shell=` to be used as `--shell=json`? Kinda ugly naming but does not break and does not introduce another API complexity?

src/commands/env.rs
```diff
@@ -45,6 +48,17 @@ fn make_symlink(config: &FnmConfig) -> Result<std::path::PathBuf, Error> {
...
+#[derive(serde::Serialize, serde::Deserialize)]
+#[allow(non_snake_case)]
+struct EnvVars {
+    FNM_MULTISHELL_PATH: String,
+    FNM_VERSION_FILE_STRATEGY: String,
+    FNM_DIR: String,
+    FNM_LOGLEVEL: String,
+    FNM_NODE_DIST_MIRROR: String,
+    FNM_ARCH: String,
+}
```
Schniz, line 60: we can probably use `std::collections::HashMap::<String, String>` as it implements Serialize

```suggestion
type EnvVars = std::collections::HashMap::<String, String>
```

src/commands/env.rs
```diff
@@ -59,50 +73,46 @@ impl Command for Env {
...
-            println!("{}", v);
+
+        let env_vars = EnvVars{
+            FNM_MULTISHELL_PATH: multishell_path.to_str().unwrap().to_owned(),
+            FNM_VERSION_FILE_STRATEGY: config.version_file_strategy().as_str().to_owned(),
+            FNM_DIR: config.base_dir_with_default().to_str().unwrap().to_owned(),
+            FNM_LOGLEVEL: <&'static str>::from(config.log_level().clone()).to_owned(),
+            FNM_NODE_DIST_MIRROR: config.node_dist_mirror.as_str().to_owned(),
+            FNM_ARCH: config.arch.to_string(),
+        };
```
Schniz, line 90: this is not formatted correctly

src/commands/env.rs
```diff
@@ -59,50 +73,46 @@ impl Command for Env {
...
+                .or_else(&infer_shell)
+                .ok_or(Error::CantInferShell)?;
+            println!("{}", shell.path(&binary_path)?);
+
+            println!("{}", shell.set_env_var("FNM_MULTISHELL_PATH", &env_vars.FNM_MULTISHELL_PATH));
+            println!("{}", shell.set_env_var("FNM_VERSION_FILE_STRATEGY", &env_vars.FNM_VERSION_FILE_STRATEGY));
+            println!("{}", shell.set_env_var("FNM_DIR", &env_vars.FNM_DIR));
+            println!("{}", shell.set_env_var("FNM_LOGLEVEL", &env_vars.FNM_LOGLEVEL));
+            println!("{}", shell.set_env_var("FNM_NODE_DIST_MIRROR", &env_vars.FNM_NODE_DIST_MIRROR));
+            println!("{}", shell.set_env_var("FNM_ARCH", &env_vars.FNM_ARCH));
```
Schniz, line 106: if `EnvVars` is `HashMap`, we can iterate over the entries and print them one by one, something like

```rust
for (env_var, value) in env_vars.iter() {
    println!("{}", shell.set_env_var(env_var, value));
}
```

Schniz (commented):
I'm having some refactoring questions/suggestions here that I would love to know your input on. But on the other side, I wonder: should we have a support for `--shell=nu[shell]`? the `json` implementation can allow implementing this on userland easier, but maybe we should bake this knowledge into fnm and test it properly to make other features like `use-on-cd` work?

Schniz:
ah and by the way, thanks for stepping up and making a PR. I really appreciate your contribution. It's not something I take for granted :pray:

src/commands/env.rs
Schniz: I know that HashMap::<String,String> is a very not type-safe thing to do but it's merely an output format that will be declared in a single place within the entire system, so I am not sure we should care 🤷

zaucy:
Thanks for the review and improvement suggestions! I'll be honest my decision making was mainly driven by my lack of experience with rust haha.

I 100% agree adding nushell to the `--shell` option would be much better. I initially went down that route, but got a little stuck with functions such as `to_clap_shell` since nushell wasn't part of the result enum. I figured the json way was less intrusive in the codebase as a first step considering my lack of rust experience.

src/commands/env.rs
zaucy: Ah I didn't know about `conflicts_with`. Thanks!

Haha I _almost_ did `--shell=json`. Maybe deprecating `--shell` and replace it with `--output`? That reads well in my head. 

```sh
fnm env --output=bash
```

Then for compatbility sake just have `--shell` map to `--output`? Is there an option for that in `clap`?

src/commands/env.rs
zaucy: Updated to use a `HashMap` now!

Schniz:
re: `to_clap_shell`:

* it's currently not returning `Option<Shell>` but I think we should do that, in the meantime we can panic like we do in WinCmd: https://github.com/Schniz/fnm/blob/c85ff97f1404b0042f7378acb0ad67b3fd607189/src/shell/windows_cmd/mod.rs#L8-L11
* we can also link to this issue: https://github.com/clap-rs/clap/issues/2778

(off topic, I'm using lists on github any time I want to link to an issue just to have the nice formatting they do on lists. FORMATTING ISSUES EVERYWHERE IN GITHUB WEN)

Schniz:
> my decision making was mainly driven by my lack of experience with rust

I built fnm in Rust when I had zero experience in Rust, I want to refactor so much of the current code but I have no time 😜 you're doing good :pray:

Schniz:
@zaucy I took the liberty to fix conflicts so I'll merge this finally. Sorry it took so long. <picture data-single-emoji=":pray_parrot:" title=":pray_parrot:">:pray_parrot:</picture>

zaucy:
Ayy! Love that

PR #839, opened 2022-10-21
Support `engines` field when `.nvmrc` / `.node-version` are not present

amitdahan:
# Summary

Adds support for `engines` field in `package.json` so having the following is supported with `fnm use` and `fnm install`:
```json
{
  "name": "my-thing",
  "engines": {
    "node": ">= 13.3.7 < 14.0.0"
  }
}
```

~~(Relies on #816 and #991)~~

Since this isn't directly equivalent to the "locked" nature of `.nvmrc` or `.node-version`, this feature will start as experimental, and behind an opt-in `--resolve-engines`:
image


ljharb:
The engines field is a range, not a single version, and affects consumers - it’s not meant to be used for the developer’s node version.

If the feature only worked when the app is private: true, then at least that would avoid the risk of breaking consumers, but it’s still a conceptual mismatch to use a ranged field for a single version.

Schniz (commented):
lgtm overall. What do you think of my comments?

kud:
That's really great. Is there any way to auto-detect this? My shell already starts `fnm use` when a config file is found.

Schniz:
fnm env --use-on-cd works with this feature for auto detection

Schniz:
If not, let us know

Schniz:
yeah seems like it's broken on `use-on-cd`, let's open an issue and fix it

Schniz:
it failed when I tested on a docker container, yet worked on my machine: it's because I'm using `--version-file-strategy=recursive`. I wonder if this should just become the default now. cc @amitdahan

gotounix:
I found it not work on Windows (PowerShell):

```
$env:FNM_MULTISHELL_PATH = "C:\Users\Max\AppData\Local\fnm_multishells\7196_1708612586200"
$env:FNM_VERSION_FILE_STRATEGY = "local"
$env:FNM_COREPACK_ENABLED = "false"
$env:FNM_RESOLVE_ENGINES = "true"
$env:FNM_DIR = "C:\Users\Max\AppData\Roaming\fnm"
$env:FNM_NODE_DIST_MIRROR = "https://nodejs.org/dist"
$env:FNM_ARCH = "x64"
$env:FNM_LOGLEVEL = "info"
E:\test [master ≡ +1 ~1 -0 !]> fnm use --resolve-engines
error: Can't find version in dotfiles. Please provide a version manually to the command.
```

@Schniz

Issue #797, opened 2022-08-13
add `fnm install --latest` to  Install latest version


lontten:

Schniz:
I think this was asked before. so might be a duplicate.
anyway, in the meantime you can use `fnm install --lts`.

ajitzero:
I think `--latest` (if added) shouldn't be same as `--lts`. Instead I suggest it should install the "current" version, i.e., the next "to-be-stable" version. Now, that would point to `18.8.0`

satvikpendem:
Agreed, I want to install the latest version that's on the Nodejs website, not the latest LTS.

ohana54:
@Schniz do you agree to the feature? I'd love to try and implement it

bnb:
I would really like to see this @Schniz. Just got a friend onboarded and they were asking how to do this, and it's something I tried to do when I started as well.

Schniz:
merged in #859, will be released soon! <picture data-single-emoji=":pray_parrot:" title=":pray_parrot:"><img class="emoji" width="20" height="auto" src="https://single-emoji.vercel.app/api/emoji/eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0..oglO8mo5T9Yg4S7e.bEzemEawvo79TxjDrqrRALHadGyFOyzMEebvjpCyBC_gm0YCC3RXSdaGBTxPvP3yB7HhVVobgjm6pVtefX6m7pfXvXJmEd9keNXuvm0SPaSayNoQsSowty1jlV0.2075YAOwPAUqPGKv0vIzuQ" alt=":pray_parrot:" align="absmiddle"></picture> 

Thanks @nzhl <picture data-single-emoji=":blob_thanks:" title=":blob_thanks:"><img class="emoji" width="20" height="auto" src="https://single-emoji.vercel.app/api/emoji/eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0..MVq-Ea4GKKKgEXz3.kBLA3XBz59UScW84ckfFNZejp9cbck6t7npRxVNfAEYSYBtmybGQqrOjw-pm54YNK3pgGFCyfIQaY4r0UZrLZRNz29fMnxxNkz5tqJhWV1us_Fvsm8Zxrn4-yiHy.iN3vt-IUsWTPShFIkbav_w" alt=":blob_thanks:" align="absmiddle"></picture>

PR #859, opened 2022-11-17
feat: support `fnm install --latest`

nzhl:
This should help to support install the latest node version using `fnm install --latest`
Just as  https://github.com/Schniz/fnm/issues/797 expected.


src/commands/install.rs
```diff
@@ -153,6 +175,8 @@ pub enum Error {
     CantFindNodeVersion { requested_version: UserVersion },
     #[error("Can't find relevant LTS named {}", lts_type)]
     CantFindRelevantLts { lts_type: crate::lts::LtsType },
+    #[error("Can't find latest version")]
+    CantFindLatest,
```
Schniz, line 179: I wonder if we should be more explicit about the issue, like: "can't find any versions in the upstream version index"

Schniz (commented):
Can you please run `pnpm generate-command-docs` to update the `commands.md` file?

I also had a small suggestion for the error message.

other than that, <picture data-single-emoji=":lgtm:" title=":lgtm:">:lgtm:</picture>

I wonder if we can add a test for it somehow. Given that the version is dynamic it might be problematic, but maybe just a smoke test with no expectations other than "not fail"?

Schniz:
Also, thank you very much for stepping in and contributing! <picture data-single-emoji=":clap_sunglasses_+1:" title=":clap_sunglasses_+1:">:clap_sunglasses_+1:</picture> 

can you run `cargo fmt` and fix `cargo clippy` issues? <picture data-single-emoji=":pray_parrot:" title=":pray_parrot:">:pray_parrot:</picture>

src/commands/install.rs
nzhl: resolved, thx for suggestion

nzhl:
> Also, thank you very much for stepping in and contributing! :clap_sunglasses_+1:
> 
> can you run `cargo fmt` and fix `cargo clippy` issues? :pray_parrot:


Thx for suggestion,  i have added smoke test and run `cargo fmt` and `cargo clippy`.

Schniz:
The commands file is not up to date because of the version change :( can you rerun the script?

nzhl:
> change

sure

nzhl:
> The commands file is not up to date because of the version change :( can you rerun the script?

done

nzhl:
> The commands file is not up to date because of the version change :( can you rerun the script?

sorry, one more question. Do i need to bump version manually ?

docs/commands.md
```diff
@@ -1,7 +1,7 @@
 # `fnm`
 
 ```
-fnm 1.32.0
+fnm 1.31.1
```
Schniz, line 4: I don’t understand why this is not updated, if you’re rebased from main and ran the script. Maybe something is funky there and I’ll fix it?

nzhl: I think it's updated rn after i run it one more time.

Schniz:
Amazing! Thank you so much.

PR #865, opened 2022-11-19
experimental: clean multishells on shell exit

Schniz:
This PR adds `--delete-on-exit` (name TBD) to allow to remove the symlinks from the multishells directory.
This was not a concern for me but I see many users that complain about it. It does introduce some complexity that I am not sure about tbh.


Schniz:
the main issue with this solution is that it will only work if the shell was closed cleanly. I believe this won't be the case most of the time.

Maybe another way we can do it is by outputting the PID as a part of the multishell, in the `fnm env`, making it more dynamic. Unfortunately that will make `--json` fail <picture data-single-emoji=":thinkies:" title=":thinkies:">:thinkies:</picture> 

another option is to make the setup be something like `fnm env --pid=$$` if you want to be able to delete it.

bah

no good option <picture data-single-emoji=":sad-t-rex:" title=":sad-t-rex:">:sad-t-rex:</picture>

PR #389, opened 2021-02-09
Chore - Use a tier 1 target for Linux binary

kaioduarte:
Accordingly to [Rust docs](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2), the previous target (`x86_64-unknown-linux-musl`, tier 2) used is not guaranteed to produce a working build.

Fix #371

PR #882, opened 2022-12-13
Build for gnu instead of musl

### LINKED ISSUE #520: Connection errors
I have similar problem:

>
> ➜  ~ fnm list-remote
> error: HttpError: error sending request for url (https://nodejs.org/dist/index.json): error trying to connect: dns error: failed to lookup address information: Try again
> ➜  ~ fnm install 14
> error: Having a hard time listing the remote versions: error sending request for url (https://nodejs.org/dist/index.json): error trying to connect: dns error: failed to lookup address information: Try again
>

Maybe it uses wrong protocol or something. `curl https://nodejs.org/dist/index.json` got the output right away.

Installed on Termux (Android 11). Shell: zsh

t-botz:
Redoing #389 which was cancelled by #554 and reintroduced the error.

Fixes #520

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -238,23 +238,17 @@ jobs:
     - uses: hecrj/setup-rust-action@v1
       with:
         rust-version: stable
-        targets: x86_64-unknown-linux-musl
+        targets: x86_64-unknown-linux-gnu
     - uses: Swatinem/rust-cache@v2
       with:
         key: static-linux-binary
-    - name: Install musl tools
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y --no-install-recommends musl-tools
     - uses: actions/checkout@v3
     - name: Build release binary
-      run: cargo build --release --target x86_64-unknown-linux-musl
-    - name: Strip binary from debug symbols
-      run: strip target/x86_64-unknown-linux-musl/release/fnm
+      run: cargo build --release --target x86_64-unknown-linux-gnu
     - uses: actions/upload-artifact@v3
       with:
         name: fnm-linux
-        path: target/x86_64-unknown-linux-musl/release/fnm
+        path: target/x86_64-unknown-linux-gnu/release/fnm
 
   build_static_arm_binary:
     name: "Build ARM binary"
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -238,23 +238,17 @@ jobs:
...
         key: static-linux-binary
-    - name: Install musl tools
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y --no-install-recommends musl-tools
     - uses: actions/checkout@v3
     - name: Build release binary
-      run: cargo build --release --target x86_64-unknown-linux-musl
-    - name: Strip binary from debug symbols
-      run: strip target/x86_64-unknown-linux-musl/release/fnm
```
t-botz, line 253: no need for strip anymore [since Cargo 1.59](https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html#creating-stripped-binaries)

t-botz:
Another small win with this PR is that the binary goes from 7.2MB to 6.5MB...
IMO there's very little reason to build a static binary with musl. fnm is meant for a dev environment supporting node, I can't see how such environment would be missing libc...

Schniz:
This means vc alpine-based docker images will fail to use fnm. Right?
Can we _add_ a gnu build? Then download it conditionally on the install script if it’s not an alpine Linux?
I wonder if we should rethink the build process and copy from Starship: https://github.com/starship/starship/blob/master/.github/workflows/release.yml

Schniz:
Ahh and also thank you very much for digging deep and figuring out the issue. Reaaaaally appreciate it 🙌

t-botz:
> This means vc alpine-based docker images will fail to use fnm. Right?

Thats right, I am not sure who does that (fnm in a docker image kind of defy the purpose IMO) but I guess if you ask, you might have encountered the use case.

> Can we add a gnu build? Then download it conditionally on the install script if it’s not an alpine Linux?
> I wonder if we should rethink the build process and copy from Starship

You mean using a matrix and `cross`? Seems like a good idea to me and fairly standard.
For the install script may be do it based on `ldd --version | grep GLIBC`? If no ldd or no GLIBC then use musl build?

I can definitely work on that.

PR #914, opened 2023-01-17
Update fish shell setup

alinpr18:


README.md
```diff
@@ -154,7 +154,8 @@ eval "$(fnm env --use-on-cd)"
 Create `~/.config/fish/conf.d/fnm.fish` add this line to it:
 
 ```fish
-fnm env --use-on-cd | source
+set PATH "$HOME/.local/share/fnm" $PATH
```
Schniz, line 157: The other shell setup steps don’t tell you to set the PATH and assume fnm is there. If you don’t use the installer you can put the binary anywhere you want so I’m not sure we need to have this line

alinpr18: Yes, what you say is correct, I already tried everything, I close this pr.

Issue #895, opened 2022-12-24
Cannot find `npm` in Git Bash for Windows


dcchuck:
* Installed FNM via chocolatey
* Added to Powershell Profile - I am able to use node and npm from Powershell
* Added to `.bashrc` for use in Git Bash

From Git Bash, I can use `node`, but when I call `npm` I see the following error -

```
chuck@MyMachine MINGW64 ~
$ npm
node:internal/modules/cjs/loader:998
  throw err;
  ^

Error: Cannot find module 'C:\Program Files\Git\Users\chuck\AppData\Local\fnm_multishells\7488_1671894125497\node_modules\npm\bin\npm-cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    at Module._load (node:internal/modules/cjs/loader:841:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.12.1
```

The path it's looking for `npm-cli.js` doesn't exist...should be pointing AppData\Local, for some reason it's looking for that under Program Files?

Is there a configuration I am missing?

Thank you

alshdavid:
Might be related to path separators.

![image](https://user-images.githubusercontent.com/12656294/214779702-2ee32b15-2113-4b9f-8362-e33352b5fbd7.png)

dcchuck:
See [issue 390](https://github.com/Schniz/fnm/issues/390) for a `.bashrc` setup that works

pierreh:
FNM_MULTISHELL_PATH wasn't fixed with this.

alshdavid:
I always install fnm manually on Windows. Download the latest release and unzip it to:

C:/Users/username/.local/fnm/bin

And add that to Windows PATH and the FNM_DIR is

C:/Users/username/.local/fnm/fnm_dir

For powershell I add the init script to $profile and git bash I add it to .bashrc (though I stopped using git bash because filepaths are always broken all the time).

Oh-my-posh does the trick and for everything else, there's Linux

PR #921, opened 2023-01-26
add Git Bash info to README

dcchuck:
I had opened an [issue](https://github.com/Schniz/fnm/issues/895) for this, eventually went with WSL2. But after the comment in there figured good to add to docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,19 @@ call "%CMDER_ROOT%\bin\fnm_init.cmd"
 ```
 You can replace `%CMDER_ROOT%` with any other convenient path too.
 
+#### Usage with Git Bash
+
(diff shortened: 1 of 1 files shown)


ModyQyW:
Just a note. `eval "$(fnm env | sed 1d)"` works for me. But `eval $(fnm env | sed 1d)` doesn't.

dcchuck:
> Just a note. `eval "$(fnm env | sed 1d)"` works for me. But `eval $(fnm env | sed 1d)` doesn't. Thanks for the PR! :D

Thank you! I've updated the PR to include quotes.

Worked on my machine, but that appears to be a quirk. Docs say to use quotes to avoid such inconsistencies 👍 [docs](https://man7.org/linux/man-pages/man1/eval.1p.html)

Schniz:
Seems like this is fixed by fnm automatically running cygwin for you in #960  🙏

PR #948, opened 2023-04-12
Add winget as installation method on Windows

fadihania:
This tool is already available on Windows Package Manager (winget). winget ships by default with Windows 11 and it would easier to use winget for installing fnm


MrSrv7:
> This tool is already available on Windows Package Manager (winget). winget ships by default with Windows 11 and it would easier to use winget for installing fnm

Yes. This is missing in the documentation. So, I thought of raising the same

MrSrv7 (approved):

PR #958, opened 2023-05-08
Fixes issue where fnm print the node version string when using fish shell in non-interactive mode

eelcocramer:
This week I ran into an [issue with `Neovide`](https://github.com/neovide/neovide/issues/1858), a neovim gui (also written in rust).

I proposed GH-955 but did not feel this was the right fix for this issue, so this PR replaces that.

Neovide starts neovim in a sub-shell and uses the neovim --embed mode to read and write msgpack messages to and from stdout.

When I start Neovide in a folder with a `.nvmrc` that defines a node version that is not my default, `fnm` prints a message and that ends up in the communication between Neovide and nvim and makes it crash.

Neovide runs a non-interactive shell, and in that mode your shell init scripts are not supposed to print anything. In that mode, the shell typically don't ready `.bashrc` and similar. However, for the `fish` shell, it seems like it always reads config.fish (https://github.com/fish-shell/fish-shell/issues/5394).

This PR adds changes the behaviour of `fnm env --use-on-cd` for the `fish` shell to add a conditional check whether the shell is interactive or not. If `fish` is in non-interactive mode `fnm use` is started with a new switch `--silent` that will not let `fnm` output the version string even if the version will change.

### DIFF
--- a/.changeset/unlucky-sheep-scream.md
+++ b/.changeset/unlucky-sheep-scream.md
@@ -1,0 +1,5 @@
+---
+"fnm": minor
+---
+
+Adds switch to silently switch node versions and uses that switch in the fish shell environment command when fish is used non-interactively.
--- a/src/commands/use.rs
+++ b/src/commands/use.rs
@@ -24,6 +24,11 @@ pub struct Use {
     /// if it will not change due to execution of this command
     #[clap(long)]
     silent_if_unchanged: bool,
+
+    /// Don't output a message identifying the version being used
+    /// even if it will change due to execution of this command
+    #[clap(long)]
+    silent: bool,
 }
 
 impl Command for Use {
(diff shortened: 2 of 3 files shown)


Schniz:
There is log level configuration that can be set to error, which will do exactly what you want, or am I missing something?

eelcocramer:
Maybe. But then the log level should be set to error as a default in non-interactive mode right?

Schniz:
Hmmm…. Define non interactive? fnm is not a shell. It’s not a debug statement, it’s an info/stdout kind of thing. I don’t think fnm should be quiet by default

Schniz:
any news on the `log-level` usage? <picture data-single-emoji=":frog-eyes:" title=":frog-eyes:">:frog-eyes:</picture>

eelcocramer:
Sorry for not getting back sooner.

I've just tried to reproduce the original issue but seems to be gone. So I guess this PR is not needed (at least not to solve the original issue).

Schniz:
<picture data-single-emoji=":ack:" title=":ack:">:ack:</picture>

thank you for the contribution <picture data-single-emoji=":meow_pensive-pray:" title=":meow_pensive-pray:">:meow_pensive-pray:</picture>

PR #963, opened 2023-05-22
Capitalised "n" to reflect default option

Joshuahuahua:
I was using this for the first time and it confused me a little, as id expect one option to be capitalized to hint to the user which is default.

Code_OS9m7gR8Jd


Schniz:
Thank you!

Schniz:
Oh, there’s no changeset :(

PR #973, opened 2023-06-28
Replace atty with std::io::IsTerminal

tottoto:
According to the GitHub Action config to build `fnm`, the stable version of Rust seems to be used. Therefore [`std::io::IsTerminal`](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html) might be able to be used, which was stabilized at 1.70.


Schniz:
<picture data-single-emoji=":noice:" title=":noice:">:noice:</picture>

PR #1018, opened 2023-07-26
Add powershell script for installation on windows

Yakiyo:
This adds a powershell script for installing fnm on windows. The ps script tries to follow the bash script as much as possible. Supported flags are:
- -version | -v for setting version (string)
- -skipShell | -s for skipping shell setup (bool)
- -installDir for setting installation dir (string)

Using only one `-` seems odd to me but thats how powershell parses arguments. As far as i've tested, it even works with only the first argument too, as in `-version`, `-v` and `-V` all do the same thing. 

To install fnm via the script, users would do
```powershell
$ irm url/to/the/script | iex
```
The bash script uses the `install` endpoint, the ps script could do ig `https://fnm.vercel.app/install.ps` or something. 

I've tested the script a few times and ensured that it works.

### DIFF
--- a/.ci/install.ps1
+++ b/.ci/install.ps1
@@ -1,0 +1,46 @@
+#!/usr/bin/env powershell
+
+param(
+  [string]$version = "latest",
+  [string]$installDir = "$env:USERPROFILE\.fnm",
+  [switch]$skipShell = $false
+)
+
+$ErrorActionPreference = 'Stop'
+
+# if version is not latest
+if (!$version.Equals("latest")) {
+  # and it does not start with a v
+  if (!$version.StartsWith("v")) {
+    # append a v in front
+    $version = "v$version"
+  }
+}
+$Target = 'fnm-windows.zip'
+
+$Url = if ($version.Equals("latest")) {
+  "https://github.com/Schniz/fnm/releases/latest/download/$Target"
+} else {
+  "https://github.com/Schniz/fnm/releases/download/$version/$Target"
+}
+
+# If installation dir does not exist, create it
+if (!(Test-Path $installDir)) {
+  New-Item $installDir -ItemType Directory | Out-Nul
+}
+
+Write-Output "Downloading from $Url"
+
+Invoke-RestMethod $Url -OutFile "$installDir\fnm.zip"
+
+Write-Output "Extracting file to $installDir"
+
+Expand-Archive "$installDir\fnm.zip" -DestinationPath "$installDir" -Force
+
+# Delete the zip file
+Remove-Item "$installDir\fnm.zip"
+
+if (!$skipShell) {
+  Write-Output "Appending 'fnm env --use-on-cd | Out-String | Invoke-Expression' to profile"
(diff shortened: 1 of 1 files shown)


Schniz:
First of all, I'd like to thank you for the time and effort and making the hard thing possible and contributing!

However, in my opinion, any installation script is a toll and out of this project scope, and I originally built the installation script to ease the installation process and it only makes sense for Linux nowadays. Since Windows has WinGet baked in as a standard installation mechanism, I don't really see the point in having a custom PowerShell installer. Do you? can you please explain why should we have an installation script and not offload to WinGet?

Yakiyo:
ig stuff like scoop and winget does exist which allows people to install fnm. But i dont think everyone uses them. Winget is certainly popular but its yet to be used by everyone, while powershell is always there in windows and a one liner script can just be used to quickly do installations. package managers are great but you wouldnt install one just to get a single app. the script itself wasnt much work, if u wanna use it, sure. if not then its all good. 
tldr theres no disadvantages but nor too many advantages either. good when u need it done quick, thats all.

beegotsy:
> can you please explain why should we have an installation script and not offload to WinGet?

Hi, I would like to install fnm under `%HOMEPATH%/.fnm` and node versions under `%HOMEPATH%/.fnm/versions` (or something like that). `-installDir` would allow me to do that, but winget doesn't.

Also, since the bash script install fnm under `$HOME/.fnm` I don't understand why on Windows in installed under `%APPDATA%/fnm`.
It should be, by default, under `%HOMEPATH%/.fnm`.
The homepath approach is an approach also used by [pyenv-win](https://github.com/pyenv-win/pyenv-win).

Am I missing something?

alexeyten:
> Also, since the bash script install fnm under `$HOME/.fnm` I don't understand why on Windows in installed under `%APPDATA%/fnm`. It should be, by default, under `%HOMEPATH%/.fnm`. The homepath approach is an approach also used by [pyenv-win](https://github.com/pyenv-win/pyenv-win).

Actually default was changed a while ago and now bash script prefers `$XDG_DATA_HOME/fnm` (which is roughly an equivalent to `%APPDATA%/fnm`) for fresh installations. It uses `$HOME/.fnm` for backward compatibility only if the directory already exists. 

See https://github.com/Schniz/fnm/blob/master/.ci/install.sh#L8-L16

beegotsy:
> Actually default was changed a while ago and now bash script prefers $XDG_DATA_HOME/fnm (which is roughly an equivalent to %APPDATA%/fnm) for fresh installations.

Okay, now it is clear; I based my previous sentence on the README file without opening the actual scripts.
Thanks for your reply.

avivkeller:
👋 Hey @Schniz! The Node.js Website Team has been recommending fnm as an official installation method. We’ve gotten some feedback around the developer experience, though. Including an installation script like this could really help smooth things out.

It would be especially helpful if the WinGet package pointed to a script like this—bundling in the extra setup steps would make the whole process much more seamless.

Issue #949, opened 2023-04-19
Feature Request: Show a progress bar when downloading


Nugine:
It's useful when the download time exceeds 5 seconds.

There is a progress bar library: https://docs.rs/indicatif

xialvjun:
really need it. In china, I don't know if it's nodejs.org is blocked or just my network speed is slow.

Schniz:
Closing this — we didn't get a follow-up. If you're still interested in this, feel free to reopen.

---

*This message was written by AI. Sorry it's not personal, but it's helping me maintain this project. Feel free to disagree with the statement and reopen the issue if needed.*

PR #1028, opened 2023-08-19
Add progress bar to install command

eblocha:
This adds the feature requested in #949. Doesn't display if the response from the download mirror doesn't have a content-length header.

I hard-coded a bar template, looks like this:

image

I think we could add an env var to customize this, but we'd need to handle syntax errors gracefully (show warning, maybe?).

Added a `--no-progress` option to turn this off. It writes to stderr.

### DIFF
--- a/src/commands/install.rs
+++ b/src/commands/install.rs
@@ -25,6 +25,10 @@ pub struct Install {
     /// Install latest version
     #[clap(long, conflicts_with_all = &["version", "lts"])]
     pub latest: bool,
+
+    /// Do not display a progress bar
+    #[clap(long)]
+    pub no_progress: bool,
 }
 
 impl Install {
@@ -34,16 +38,19 @@ impl Install {
                 version: v,
                 lts: false,
                 latest: false,
+                no_progress: _,
             } => Ok(v),
             Self {
                 version: None,
                 lts: true,
                 latest: false,
+                no_progress: _,
             } => Ok(Some(UserVersion::Full(Version::Lts(LtsType::Latest)))),
             Self {
                 version: None,
                 lts: false,
                 latest: true,
+                no_progress: _,
             } => Ok(Some(UserVersion::Full(Version::Latest))),
             _ => Err(Error::TooManyVersionsProvided),
         }
@@ -55,6 +62,8 @@ impl Command for Install {
 
     fn apply(self, config: &FnmConfig) -> Result<(), Self::Error> {
         let current_dir = std::env::current_dir().unwrap();
+
+        let show_progress = !self.no_progress;
 
         let current_version = self
             .version()?
@@ -131,6 +140,7 @@ impl Command for Install {
             &config.node_dist_mirror,
             config.installations_dir(),
             safe_arch,
+            show_progress,
         ) {
             Err(err @ DownloaderError::VersionAlreadyInstalled { .. }) => {
                 outln!(config, Error, "{} {}", "warning:".bold().yellow(), err);
@@ -225,6 +235,7 @@ mod tests {
             version: UserVersion::from_str("12.0.0").ok(),
             lts: false,
             latest: false,
+            no_progress: true,
         }
         .apply(&config)
         .expect("Can't install");
@@ -250,6 +261,7 @@ mod tests {
             version: None,
             lts: false,
             latest: true,
+            no_progress: true,
         }
         .apply(&config)
         .expect("Can't install");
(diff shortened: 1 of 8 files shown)


src/progress.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        if let Some(ref bar) = self.progress {
+            bar.inc(size as u64);
+        }
+
+        Ok(size)
+    }
+}
+
+impl Drop for ResponseProgress {
```
eblocha, line 51: Using Drop because the XzDecoder takes ownership of the reader. Ensures the progress bar is finalized once the download is completed. Not sure if there's a better way of doing this.

src/progress.rs
```diff
@@ -0,0 +1,55 @@
...
+    response: Response,
+}
+
+fn make_progress_bar(size: u64) -> ProgressBar {
+    let bar = ProgressBar::new(size);
+
+    bar.set_style(
+        ProgressStyle::default_bar()
+            .template("[{elapsed_precise}] [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, {eta})")
+            .unwrap()
```
eblocha, line 17: This can return a `TemplateError` from indicatif when the template has a syntax issue. Unwrapping because it's a static template.

src/commands/install.rs
```diff
@@ -25,6 +25,10 @@ pub struct Install {
     /// Install latest version
     #[clap(long, conflicts_with_all = &["version", "lts"])]
     pub latest: bool,
+
+    /// Do not display a progress bar
+    #[clap(long)]
+    pub no_progress: bool,
```
eblocha, line 31: I don't think clap has a way to add `--progress/--no-progress` to enable or disable a flag, just made a progress bar the default.

src/downloader.rs
```diff
@@ -63,10 +65,7 @@ fn download_url(base_url: &Url, version: &Version, arch: &Arch) -> Url {
     .unwrap()
 }
 
-pub fn extract_archive_into<P: AsRef<Path>>(
-    path: P,
-    response: crate::http::Response,
-) -> Result<(), Error> {
+fn extract_archive_into<P: AsRef<Path>, R: Read>(path: P, response: R) -> Result<(), Error> {
```
Schniz, line 68: <picture data-single-emoji=":interesting:" title=":interesting:">:interesting:</picture>

Schniz: just a thought (not important, can be done in a different PR): what do you think about starting to use `impl` instead of generic constraints? seems more readable right?

```suggestion
fn extract_archive_into(path: impl AsRef<Path>, response: impl Read) -> Result<(), Error> {
```

Schniz (approved):
<picture data-single-emoji=":lgtm:" title=":lgtm:">:lgtm:</picture> <picture data-single-emoji=":frog-10:" title=":frog-10:">:frog-10:</picture> wdyt about my small comment?

In a future PR it would be cool to add the extraction into account too, so we have a multi progress bar if necessary to show download+extraction which are both tasks that stream and take time 😌

src/progress.rs
```diff
@@ -0,0 +1,56 @@
...
+            "[{elapsed_precise}] [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, {eta})",
+        )
+        .unwrap()
+        .progress_chars("#>-"),
+    );
+
+    bar
+}
+
+impl ResponseProgress {
```
Schniz, line 25: you think we should unit test this somehow?

eblocha: I added a test to ensure we still read from the underlying response with the progress wrapper. Also added a `indicatif::TermLike` impl that captures the progress bar activity and saves all the activity into a String, then checks for "#" characters in the string after reading all the data.

The mock impl is a bit verbose so I added it in a separate commit to revert it easily. Let me know if you'd prefer to just test the read operations pass through.

eblocha:
Looks like there are some failing e2e tests, saying "operation timed out" when trying to hit the proxy. Any ideas what's going on there? Some kind of rate limiting from the nodejs servers?

bharat5604 (approved):

Schniz:
I'll merge this and add some fixes, thank you so much for this contribution and sorry I haven't merged it till now <picture data-single-emoji=":meow_pensive-pray:" title=":meow_pensive-pray:">:meow_pensive-pray:</picture>

src/commands/install.rs
Schniz: fyi, I'm gonna change that before releasing to `--progress=never`, and will have `--progress=auto|always` so `auto` will be based on the log level <picture data-single-emoji=":blob_thumbs_up:" title=":blob_thumbs_up:">:blob_thumbs_up:</picture>

PR #1034, opened 2023-08-31
modify cmd command

pping:
set up a startup script](https://superuser.com/a/144348) and append the following line does not work :

FOR /f "tokens=*" %i IN ('fnm env --use-on-cd') DO CALL %i
After my testing, this will cause a endless loop,  causing the cmd to not function properly.

add follwing command:
::autorun.cmd
fnm env > %USERPROFILE%/fnm.txt
for /f "delims=" %%a in (%USERPROFILE%/fnm.txt) do (
    call %%a
)
del %USERPROFILE%/fnm.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,22 @@ fnm env --use-on-cd | Out-String | Invoke-Expression
 
 fnm is also supported but is not entirely covered. [You can set up a startup script](https://superuser.com/a/144348) and append the following line:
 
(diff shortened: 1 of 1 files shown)


Schniz:
I don't understand this, so I'm closing. Feel free to reopen a PR with a test or a video shows a reproduction please :pray:

PR #1047, opened 2023-09-26
Simplified Windows PowerShell setup

Megamannen:
Removed section about different profile location and instead used `$profile` variable


Schniz (approved):
Pffff. Nice catch. Thank you!

Issue #681, opened 2022-02-23
Can't use system as default with --use-on-cd and --version-file-strategy recursive


NicolasWebDev:
Hi,

With `eval "$(fnm env --version-file-strategy recursive --use-on-cd)"` on my bashrc and system set as default, when opening a shell I have the following error:
```
error: Could not find any version to use. Maybe you don't have a default version set?
Try running `fnm default <VERSION>` to set one, or create a .node-version file inside your project to declare a Node.js version.
```

Creating a .node-version with "system" inside doesn't solve the problem.

If I remove one of both flags, then it works. If I set default to another version, then it works. `fnm default system` doesn't give any error. `fnm use system` either.

I am using fnm-bin 1.31.0 on arch.

sarmong:
Same problem.

aquacash5:
After looking over the code, this error seems to come from the fact that `--version-file-strategy recursive` runs on every `cd`, not just the ones that it may be useful like `--version-file-strategy local`.

Local: 

```bash
__fnm_use_if_file_found() {{
    if [[ -f .node-version || -f .nvmrc ]]; then
        fnm use --silent-if-unchanged
    fi
}}

__fnmcd() {{
    \cd "$@" || return $?
    __fnm_use_if_file_found
}}

alias cd=__fnmcd
```

Recursive:

```bash
__fnm_use_if_file_found() {{
    fnm use --silent-if-unchanged
}}

__fnmcd() {{
    \cd "$@" || return $?
    __fnm_use_if_file_found
}}

alias cd=__fnmcd
```

---

A solution is to suppress the `CantInferVersion` error if `silent_if_unchanged` is true. Instead fnm could add a debug line saying that fnm is suppressing the error due to the use of `silent_if_unchanged`. This would only be visible when the user sets the `RUST_LOG` environment variable.

Technically, if fnm does not find a version, fnm does not change the version, so fnm could silence the error with `silent_if_unchanged`.

aquacash5:
The issue is if you make your default `system`. On *nix systems, fnm sets the default alias to symlink to `/dev/null/installation`. When fnm tries to get the default version, fnm calls [`canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) on that symlink.

Since `canonicalize` follows symlinks, we get a `NotADirectory` error; Instead of, the system alias.

PR #1049, opened 2023-10-09
Add default if none option for FnmConfig

aquacash5:
Fixes: #681


Schniz:
question: doesn't that already work with the recursive file strategy? i.e. `~/Code/.node-version` with `v18.0.0` will resolve if you don't have `~/Code/my-project/.node-version`

If I'm missing anything, can you please explain this to me? 🙏

aquacash5:
Currently, yes, if, as in your example, you only change directories from `~/Code` down. However if you change to a directory where no folder up to the root has a `.node-version` you get an error. As reported in #681.

In the case where no `.node-version` or `.nvmrc` file is found in any parent directory, fnm would change the node version to whatever is the fnm default.

For example, on my computer, if I am not in a project that uses node, I want node and npm to be my system version. When I change directories from `~/Code/node-web-project` to `/etc` or `~/Documents`, the current implementation will show an error and leave the last node setup from `~/Code/node-web-project` as the setup. With this option it will change back to the `system` node setup.

A work around would be to create a `/.node-version` file with the contents of `system`, but that feels bad, littering the `/` directory with a configuration file.

aquacash5:
I see now that this is supposed to be the default behavior. I will follow up on #681.

PR #1050, opened 2023-10-12
Allow for cd /D on windows

JakeHandsome:
When installed on windows with the `--use-on-cd` flag. Using `cd /D <path>` will fail because only the `/D` argument is passed into the actual cd command.

Replacing %1 with %* will forward all the arguments so `cd /D <path>` functions.

### DIFF
--- a/.changeset/many-paws-fetch.md
+++ b/.changeset/many-paws-fetch.md
@@ -1,0 +1,5 @@
+---
+"fnm": patch
+---
+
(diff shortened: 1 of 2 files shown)


Schniz (approved):
<picture data-single-emoji=":sgtm:" title=":sgtm:">:sgtm:</picture> <picture data-single-emoji=":blob_thanks:" title=":blob_thanks:">:blob_thanks:</picture>

PR #1058, opened 2023-10-19
fix: return default version if canonicalize fails

aquacash5:
Since `canonicalize` can fail when default on *nix operating systems is set to `system` (#681), I have changed `find_default_version` to return the default alias if `canonicalize` fails.


gutenye:
@Schniz Will this get reviewed and merged?

Schniz:
Sorry for my lack of work in this regard, I'm in reserves in IDF since the massacre of Oct 7th, so have limited free time. I'll review now.

PR #1066, opened 2023-10-30
This $PATH should be wrapped in quotes so it will work with spaces

floh1695:
This $PATH use needs to be wrapped in quotes so it will expand properly if there are spaces within the $PATH. This seems to only effect sh, not bash.

To reproduce:
1. Run `PATH="space path:$PATH"`
2. Run `eval $(fnm env)`
3. At this point I would expect you to get the error: `sh: 1: export: path:rest/off/you/path/before/next/space: bad variable name`

### DIFF
--- a/.changeset/violet-rice-brake.md
+++ b/.changeset/violet-rice-brake.md
@@ -1,0 +1,5 @@
+---
+"fnm": patch
+---
+
+Fixes a bug when running `eval $(fnm env)` in sh when there a spaces in the $PATH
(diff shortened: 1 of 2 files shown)


floh1695:
The the fails for clippy and bench/linux don't seem related to what I changed. Please let me know if I'm mistaken about that.

Schniz:
Yeah that’s fine. I will also accept a `cargo clippy —fix` PR or will try doing it today. Mind pushing a changeset so it’ll appear on the changelog? (Check out the comment by the bot for help)

Schniz:
And thanks for contributing! I don’t take this lightly

floh1695:
I have pushed up a patch changeset. Is that correct or should I have made it a minor change?

And no worries, thanks for providing such a great tool!

Schniz:
Patch is good! Thank you

PR #1083, opened 2023-12-27
Add workaround for endless loop of WinCMD script

stkevintan:
I just set up the CMD auto run script in my windows machine (Windows 11 Enterprise, 23H2, x64) 
I firstly encountered `%i was unexpected at this time.`, so I changed to script the Cmder version (double the `%`)
after that, the cmd hangs forever, and a huge amount of cmd process emerged:
image

I got the answer from https://stackoverflow.com/a/57451662
> The problem is, FOR /F starts a new instace of cmd.exe and that instance also starts your auto run script, this can result into an endless loop.

So I updated the README, hope my experience can help other peopoe

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,18 @@ fnm is also supported but is not entirely covered. [You can set up a startup scr
 FOR /f "tokens=*" %i IN ('fnm env --use-on-cd') DO CALL %i
 ```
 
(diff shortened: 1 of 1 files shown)


Schniz:
I believe it was addressed in a different PR. Thank you for the contribution <picture data-single-emoji=":blob_salute:" title=":blob_salute:">:blob_salute:</picture>

PR #1095, opened 2024-02-07
chore: updates powershell to power-shell as per the CLI app

clearlysid:


VinitGurjar:
When will this issue merge? 

> I have the same problem, so I thought of raising an issue, but when I came here there were many issues  present with the same problem but none of them are merged yet.

PR #1102, opened 2024-03-13
fix: fish shell path duplicates

truehumandesign:
As fish will reappend the PATH by simply using `set -gx PATH...` you get duplicate entries if you have nested fish sessions (e.g. tmux). 
`fish_add_path -gP` ensures that the PATH is only set once and is prepended to the current PATH

references #441


truehumandesign:
Closing this as after more inspecting this would be the wrong way - every instance of fish should have its own `PATH` to the newly generated multi shell. As the newly generated one gets prepended it works as supposed I think.

PR #1109, opened 2024-04-11
fix: support x64-musl

Vinfall:
Yes, an easy fix as silly as bypassing the hardcoded check and then set `FNM_NODE_DIST_MIRROR` & `FNM_ARCH` in shell rc.


Vinfall:
I think it's also possible to support other musl architectures, if the unofficial node mirror has builds for those. But I'm lazy enough to pretend they do not exist.

PR #1112, opened 2024-04-30
fix(install.sh): support windows mingw platform

ahaoboy:
```
 uname -s
MINGW64_NT-10.0-22635
```


Schniz:
woah, that's wonderful. thank you!

PR #1143, opened 2024-05-29
Support XDG convention in MacOS too

Schniz:
This PR resolves #959 as it is based on it and I'm pushing more code to make it compatible with the changes done while it wasn't merged.

Thank you @utkarshgupta137 for the initial work :pray:


utkarshgupta137:
Looks good to me, although I'm no longer using Node.js or fnm, so I won't be able to test this.

PR #1159, opened 2024-06-02
fix 'insecure world writable dir' error from gem

### LINKED ISSUE #1158: Running `gem list --local` complains about world-writable fnm directory in $PATH
* MacOS Sonoma 14.5
* fnm 1.37.0
* gem 3.0.3.1 (default version for Sonoma 14.5)

```
$ gem list --local
/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/universal-darwin23/rbconfig.rb:21: warning: Insecure world writable dir /Users/simon/.local/state/fnm_multishells in PATH, mode 040777

*** LOCAL GEMS ***

[deleted]
```

Schniz:
fixes #1158, hopefully.
when we create the multishell path, we will ensure only the owner can read and write to it.

### DIFF
--- a/src/commands/env.rs
+++ b/src/commands/env.rs
@@ -35,17 +35,16 @@ fn generate_symlink_path() -> String {
 }
 
 fn make_symlink(config: &FnmConfig) -> Result<std::path::PathBuf, Error> {
-    let base_dir = if cfg!(windows) {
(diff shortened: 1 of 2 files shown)


Schniz:
can't repro

PR #1174, opened 2024-06-12
Update README.md - improved PowerShell instructions

Remzi1993:
You were missing some steps for people who didn't have a PowerShell profile file in the first place. I added those instructions for newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ fnm env --use-on-cd | source
 ```
 
 #### PowerShell
-First create your PowerShell profile file if you haven't:
+Create your PowerShell Profile File (if it doesn't exist):
 ```powershell
 # Check if the directory for the profile exists, if not, create it
 $profileDir = Split-Path $profile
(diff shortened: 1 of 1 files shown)


Schniz:
Can't this be something like `mkdir -p $(dirname $profile)`? Seems very excessive to have to write all this thing. I also expect that ppl will know to configure their shell or at least create directories?

Remzi1993:
> Can't this be something like `mkdir -p $(dirname $profile)`? Seems very excessive to have to write all this thing. I also expect that ppl will know to configure their shell or at least create directories?

I didn't know that and I have been using Powershell for a couple of months and never needed to. This is the first tool I encounter that I need to do it.

README.md
```diff
@@ -169,18 +169,35 @@ fnm env --use-on-cd | source
...
 ```
 
-- For macOS/Linux, the profile is located at `~/.config/powershell/Microsoft.PowerShell_profile.ps1`
-- On Windows to edit your profile you can run this in a PowerShell
+Open the Profile File in a text editor:
+- On Windows to edit your profile you can run this in PowerShell
   ```powershell
   notepad $profile
   ```
+- For macOS/Linux, the profile is located at `~/.config/powershell/Microsoft.PowerShell_profile.ps1`
```
Schniz, line 191: I think we can avoid this and simply have the following as "edit your profile" code:

```powershell
New-Item -ItemType Directory -Force -Path (Split-Path -Parent $profile)
notepad $profile
```

Remzi1993: I don't know, if you tested this then I would presume that it is better. My advice would be to test it in a virtual machine with a brand new PowerShell 5 or 7 to check if this works.

Remzi1993: @Schniz Have you look at this?

PR #1184, opened 2024-06-22
fix: list-remote --filter --latest panics when filter has no results

julescubtree:
This PR checks for results before modifying the `all_versions` vector, and opts for `.swap()` and `.truncate()` over `.drain()` to reduce overhead and improve efficiency.


Schniz (approved):
thank you
i'll add a test :)

PR #1204, opened 2024-07-27
Create docs/nightly.md

### LINKED ISSUE #610: Node 18 Nightly support

JLarky:
I wasn't sure where is the best place to put this info, so decided that a file in docs should be good enough start :)

fixes #610

### DIFF
--- a/docs/nightly.md
+++ b/docs/nightly.md
@@ -1,7 +1,7 @@
 # Working with Node nightly builds
 
 fnm doesn't give you any shorthands to install nightly builds, but you can
-do that by simply proviging `--node-dist-mirror` flag. 
+do that by manually passing a "dist mirror" with the `--node-dist-mirror` flag. 
 
 ## Example usage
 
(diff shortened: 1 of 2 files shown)


docs/nightly.md
```diff
@@ -0,0 +1,26 @@
+# Working with Node nightly builds
+
+fnm doesn't give you any shorthands to install nightly builds, but you can
+do that by simply proviging `--node-dist-mirror` flag.
```
Schniz, line 4:
```suggestion
do that by manually passing a "dist mirror" with the `--node-dist-mirror` flag. 
```

Schniz (approved):
interesting, thank you! fixed a small typo and removed usage of "simply" <picture data-single-emoji=":meow_pensive-pray:" title=":meow_pensive-pray:">:meow_pensive-pray:</picture> 
do you think we should have a `--nightly` variant as a global config? how common is it to download a nightly?

JLarky:
Thank you for the review, I applied the suggested changes.

As for how popular it is, if usage of https://www.npmjs.com/package/node-nightly shows, it's not very popular, but it's strictly more than zero :)

For me personally it would have been a bit more convenient to have --nightly in terms of discovery, but as long as there's **a** way to do it I'm fine with current state :)

PR #1222, opened 2024-08-11
Add `--shell` to `fnm env` commands in the README

mo8it:
When specifying `--shell`, the `fnm env` command takes about `1.3ms` instead of `54ms` on my machine. This is about 40x speedup.

The README has a section for every shell anyway, so adding the option saves some time on shell initialization.


Schniz (approved):
That's the spirit! Thank you.

PR #1223, opened 2024-08-11
Use `var_os` instead of `var

mo8it:
`var_os` doesn't need to check if the value of the variable is in UTF8. `split_paths` and `PathBuf` can deal with `OsString`


Schniz:
nice! can you please add a changeset? 🙏

PR #1250, opened 2024-09-12
Add Windows arm64 release binaries

### LINKED ISSUE #1068: Windows on Arm
Any plans or status on windows on arm support?

fnm itself can stay in x86 as the performance is good enough. But on windows on arm machine, trying to install node v20.x that has  official windows on arm support now, fnm would still just install the x86 version.

dennisameling:
Closes #1068 

Even though the GitHub Runner hosts are x64, MSVC and Rust support cross-compiling to arm64 perfectly fine on Windows. This PR uploads arm64 artifacts for `fnm`.

### DIFF
--- a/.changeset/heavy-seals-stare.md
+++ b/.changeset/heavy-seals-stare.md
@@ -1,0 +1,5 @@
+---
+"fnm": minor
+---
+
+feature: Add Windows arm64 release binaries
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -75,7 +75,13 @@ set_filename() {
     USE_HOMEBREW="true"
     echo "Downloading fnm using Homebrew..."
   elif [ "$OS" = "Windows" ] ; then
-    FILENAME="fnm-windows"
+    case "$(uname -m)" in
+      aarch* | armv8*)
+        FILENAME="fnm-windows-arm64"
+        ;;
+      *)
+        FILENAME="fnm-windows"
+    esac
     echo "Downloading the latest fnm binary from GitHub..."
   else
     echo "OS $OS is not supported."
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -57,16 +57,25 @@ jobs:
       - uses: hecrj/setup-rust-action@v1
         with:
           rust-version: ${{env.RUST_VERSION}}
+          targets: x86_64-pc-windows-msvc,aarch64-pc-windows-msvc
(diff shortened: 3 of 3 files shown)


dennisameling:
Successful CI run in my fork (given that the CI in this PR is pending maintainer approval): https://github.com/dennisameling/fnm/actions/runs/10829319005/job/30046612481?pr=1

This basically doubles the build time for the "Release build for Windows" job, because the builds run sequentially. Alternatively, I could change the CI config in such way that both builds (x64 and arm64) run in parallel. WDYT @Schniz?

afbeelding

Schniz:
@dennisameling will it download the correct Windows version off Node.js dist?

dennisameling:
@Schniz it will indeed! The only caveat is that official Windows arm64 support was added in Node 20 (see e.g. [here](https://nodejs.org/dist/v20.0.0/)). Windows arm64 supports x64 emulation, so folks needing older versions can always do something like `fnm install 18 --arch x64` to get those. That might be nice to add to the release notes.

I don't think `fnm` would need to implement any workarounds for that; looking at the [LTS schedule](https://nodejs.org/en/about/previous-releases#release-schedule), v18 support will end in May 2025, v20 is already LTS, and v22 will become LTS later this month as well.

Here's some examples using the `arm64` binary I built locally:

```
PS > .\fnm install 20
Installing Node v20.18.0 (arm64)
00:00:07 ██████████████████████████████████████████████████████████████████████████ 25.00 MiB/25.00 MiB (3.28 MiB/s, 0s)

PS > .\fnm install 18
Installing Node v18.20.4 (arm64)
error: Can't download the requested binary: v18.20.4 for arm64 not found upstream.
You can `fnm ls-remote` to see available versions or try a different `--arch`.
```

ahaoboy:
May also need to modify install.sh, and perhaps add binstall support at the same time

https://github.com/Schniz/fnm/issues/1280

If use these two names, you can skip the compilation using cargo binstall
```
fnm-x86_64-pc-windows-msvc.zip
fnm-arm64-pc-windows-msvc.zip
```

dennisameling:
This can potentially be superseded by https://github.com/Schniz/fnm/pull/1282

montyc1999:
Is there anything else blocking this @Schniz?

Currently, there's only an x64 binary available on winget.

It would be nice if I could just `winget install Schniz.fnm` on my win-arm64 machine, and not have to remember the `--arch` flag to install the correct node.js.

Issue #1100, opened 2024-02-28
`--use-on-cd` does not respect `--resolve-engines` with `--version-file-strategy=local` 


martonperei:
The current hooks only run if an `.nvmrc` or `.node-version` file exists. It should probably check `package.json` too, if `--resolve-engines` is set.

lobosan:
Same issue here

PR #1259, opened 2024-09-27
Fix `--resolve-engines` in combination with `--use-on-cd`

skirsten:
The shell output was not checking for the package.json file.

Fixes https://github.com/Schniz/fnm/issues/1100

### DIFF
--- a/.changeset/hot-mice-camp.md
+++ b/.changeset/hot-mice-camp.md
@@ -1,0 +1,5 @@
+---
+"fnm": patch
+---
+
+Fix `--resolve-engines` in combination with `--use-on-cd`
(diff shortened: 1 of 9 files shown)


Schniz:
oh damn i didn't see this PR and I committed #1263 with some tests. let me attribute you in that PR

Schniz:
i love your code more, i'll just add my tests to this PR 😁

skirsten:
Thanks for merging this.
I had a quick look at your code and noticed that I overlooked the `src/shell/windows_cmd/cd.cmd` file.

So I believe on Windows with CMD it still missing the fix.

PR #1263, opened 2024-10-06
    fix: package.json should be read in use-on-cd if resolve_engines is set

Schniz:
- **fix: package.json should be read in use-on-cd if resolve_engines is set**
- **apply tests**
- **add changeset**


Schniz:
#1259 is better, i'll add the e2e tests there

PR #1265, opened 2024-10-06
feat: enable --resolve-engines by default. out of experimental phase!

Schniz:
- **bump Rust to 1.81**
- **feat: enable --resolve-engines by default. out of experimental phase!**
- **add changeset**

cc @amitdahan for this wonderful feature that is more than 1yr old now

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  RUST_VERSION: "1.78"
+  RUST_VERSION: "1.81"
 
 jobs:
   fmt:
--- a/src/commands/ls_remote.rs
+++ b/src/commands/ls_remote.rs
@@ -13,7 +13,10 @@ pub struct LsRemote {
 
     /// Show only LTS versions (optionally filter by LTS codename)  
     #[arg(long)]
-    #[allow(clippy::option_option)]
+    #[expect(
+        clippy::option_option,
+        reason = "clap Option<Option<T>> supports --x and --x=value syntaxes"
+    )]
     lts: Option<Option<String>>,
 
     /// Version sorting order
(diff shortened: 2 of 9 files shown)


.github/workflows/rust.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  RUST_VERSION: "1.78"
+  RUST_VERSION: "1.81"
```
amitdahan, line 14: Related?

src/commands/ls_remote.rs
```diff
@@ -13,7 +13,10 @@ pub struct LsRemote {
 
     /// Show only LTS versions (optionally filter by LTS codename)  
     #[arg(long)]
-    #[allow(clippy::option_option)]
+    #[expect(
+        clippy::option_option,
+        reason = "clap Option<Option<T>> supports --x and --x=value syntaxes"
+    )]
```
amitdahan, line 19: Related?

src/config.rs
```diff
@@ -77,16 +77,24 @@ pub struct FnmConfig {
...
         global = true,
         hide_env_values = true,
         verbatim_doc_comment
     )]
-    resolve_engines: bool,
+    #[expect(
+        clippy::option_option,
+        reason = "clap Option<Option<T>> supports --x and --x=value syntaxes"
+    )]
+    resolve_engines: Option<Option<bool>>,
```
amitdahan, line 97: Why is it `Option<Option<bool>>` and not `Option<bool>`?

Schniz: because `Option<Option<bool>>` means that `--resolve-engines` will work and so will `--resolve-engines=false`

amitdahan: Ugh

So `--resolve-engines` is `Some(None)`, and `--resolve-engines=false` is `Some(Some(false))`?

PR #1268, opened 2024-10-06
Add `fnm default` to return default version

### LINKED ISSUE #1196: Determine fnm's default version programmatically
I use `fnm default <version>`.  I then regularly run a script to check if my default version is out of date.  However, I can't find a programmatic way to determine what fmn default is set to.  I'm using `fnm current`, but this doesn't always work, if I'm in shell or location that has an overridden version.

### Expected behavior
- `fnm default` (without further arguments) should return fnm's current default version. 
- Or alternatively, there should be documented file I can `cat` to see fnm's default version.

### Actual behavior
`fnm default` prints help, no file is documented.

isaacl:
Fixes #1196.


isaacl:
Note: my rust is pretty rusty, so check extra carefully for common rust mistakes!

isaacl:
@Schniz ping

PR #1271, opened 2024-10-07
Return x64-musl for built musl binaries

fabricionaweb:
I think it would be nice if the `FNM_ARCH` could returns `x64-musl` when is built on musl.


Schniz:
feels like we need to have a list of prioritized platforms, right? so we return `[musl, x64]` for musl builds, and `arm64, x64` for #1068 for instance. we can know which files exist based on the index, or just by requesting them one by one.

PR #1304, opened 2024-10-31
feat: add 'info' command to display current, LTS, and latest Node.js

GNURub:
Add info command: 

> $ fnm info
> Using: v23.1.0
> Current LTS: v22.11.0
> Current latest: v23.1.0


Schniz:
can we discuss about the use case?
if we go forward with this, I assume it will be `info-remote` because I want to show that it's gonna fetch data from a remote host, unlike `fnm info` that might tell you what's the current state of fnm itself.

GNURub:
> can we discuss about the use case? if we go forward with this, I assume it will be `info-remote` because I want to show that it's gonna fetch data from a remote host, unlike `fnm info` that might tell you what's the current state of fnm itself.

maybe info-node better?

PR #1334, opened 2024-11-22
Update readme to add "-" on powershell command to be the same value as what the cli requires

RodrigoScola:
the current readme command is: "fnm env --use-on-cd --shell powershell | Out-String | Invoke-Expression"

but copy pasting it show the error: 
```bash
error: invalid value 'powershell' for '--shell <SHELL>'
  [possible values: bash, zsh, fish, power-shell, cmd]

  tip: a similar value exists: 'power-shell'

For more information, try '--help'.
```

changing to "fnm env --use-on-cd --shell power-shell | Out-String | Invoke-Expression" to be able to copy and paste it

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ fnm env --use-on-cd --shell fish | source
 Add the following to the end of your profile file:
 
 ```powershell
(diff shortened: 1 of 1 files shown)


Schniz:
what version are you on? both work on fnm 1.38.1

PR #1337, opened 2024-11-29
docs(readme): improve info about powershell profile

SunsetTechuila:
New command will open the profile in the editor associated with `.ps1` files on any OS


Schniz (approved):
thank you

PR #1348, opened 2024-12-21
Update README.md

nbro:


nbro:
@fabricionaweb Well, with fnm 1.37.1, I could only use power-shell, but if with newer versions both work, feel free to close this.

PR #1353, opened 2024-12-25
fnm env --global: considering if this is a good idea

Schniz:
technically i think that i can just mutate the 'default' alias instead. because now that i think about it,
it makes 0 sense to set current to 'default' every time a new shell is opened. so current _is_ default.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -401,6 +401,9 @@ Options:
       --json
           Print JSON instead of shell commands
 
+      --global
+          Have a single Node.js version defined globally. This opts out of the "multishell" solution which allows a Node.js version per shell session. All your shell sessions will share a global Node.js version, and calling `fnm use` will update the global pointer
+
       --log-level <LOG_LEVEL>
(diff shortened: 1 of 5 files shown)


docs/commands.md
```diff
@@ -401,21 +401,24 @@ Options:
       --json
           Print JSON instead of shell commands
 
+      --global
```
Schniz, line 404: :gremlin-ohno:

Schniz, line 404: hi? :gremlin-ohno:

Schniz: :cache:

Schniz: <picture data-single-emoji=":cache:" title=":cache:">:cache:</picture>

Schniz, line 404: test?!

Schniz, line 404: <picture data-single-emoji=":gremlin-ohno:" title=":gremlin-ohno:">:gremlin-ohno:</picture>

PR #1403, opened 2025-03-27
Add instructions to setup the file for Clink

orzgithub:
Added instructions for using fnm with Clink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,40 @@ call "%CMDER_ROOT%\bin\fnm_init.cmd"
 
 You can replace `%CMDER_ROOT%` with any other convenient path too.
 
+#### Useage with Clink
+
+By default Clink load *.lua under %LOCALAPPDATA%\clink when starting a new WinCMD instence so the startup script should written in lua. So this script will convert the CMD-style environment variable setting statement to lua-style. Write the following script into %LOCALAPPDATA%\clink\fnm.lua (or other path, depending on your clink configuration):
+
+```lua
+if (clink.version_encoded or 0) < 10020030 then
+  error("fnm requires a newer version of Clink; please upgrade to Clink v1.2.30 or later.")
+end
+
+if( os.getenv(FNM_AUTORUN_GUARD) == nil )
+then
+	os.setenv('FNM_AUTORUN_GUARD', 'AutorunGuard')
+	local command_handle = io.popen('fnm env --use-on-cd', "r")
+	local command_result = command_handle:read("*a")
+	command_handle:close()
+	for line in command_result:gmatch("([^\r\n]+)") do
+		if line:sub(1,4) == "SET " or line:sub(1,4) == "set " then
+			local rest = line:sub(5)
+			local eq_pos = rest:find("=", 1, true)
+			if eq_pos then
+				local var = rest:sub(1, eq_pos - 1):match("^%s*(.-)%s*$")
+				local value = rest:sub(eq_pos + 1):match("^%s*(.-)%s*$")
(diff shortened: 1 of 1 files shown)


bugficks:
Guess should have check PRs first... this is what I came up with:
+ tries to remove symlink on exit.
+ does doskey functionality in lua (haven't tested that though)

```lua

if os.getenv("FNM_AUTORUN_GUARD") == nil then
    os.setenv("FNM_AUTORUN_GUARD", "1")

    local function parse_fnm_env()
        local handle = io.popen('fnm env --use-on-cd 2>nul')
        if not handle then return nil end
        local out = handle:read("*a")
        handle:close()

        local env = {}
        for line in out:gmatch("[^\r\n]+") do
            -- Matches: set VAR=VALUE  (quotes rarely used, but handle if present)
            local var, value = line:match("^[Ss][Ee][Tt]%s+([^=]+)=(.*)$")
            if var and value then
                value = value:gsub('^"(.*)"$', '%1') -- remove surrounding quotes if present
                env[var:match("^%s*(.-)%s*$")] = value -- trim spaces from var
            end
            -- local line = line:match("^([Dd][Oo][Ss][Kk][Ee][Yy]%s+.*)$")
            -- if line then
            --     os.execute(line)
            -- end
        end
        return env, out
    end

    local function setup_fnm()
        local env, raw = parse_fnm_env()
        if not env or not raw then return end

        for var, value in pairs(env) do
            os.setenv(var, value)
        end
    end

    setup_fnm()
end

local function check_and_use_fnm()
    -- Check FNM_VERSION_FILE_STRATEGY environment variable
    local fnm_strategy = os.getenv("FNM_VERSION_FILE_STRATEGY")
    
    -- Check if we should use recursive strategy
    if fnm_strategy == "recursive" then
        os.execute("fnm use --silent-if-unchanged")
    else
        -- Check for .nvmrc file
        local nvmrc = io.open(".nvmrc", "r")
        if nvmrc ~= nil then
            io.close(nvmrc)
            os.execute("fnm use --silent-if-unchanged")
        else
            -- Check for .node-version file
            local node_version = io.open(".node-version", "r")
            if node_version ~= nil then
                io.close(node_version)
                os.execute("fnm use --silent-if-unchanged")
            end
        end
    end
end

-- remove fnm symlinks on exit
clink.onendedit(function(line)
    if line:match("^exit") then
        local path = os.getenv("FNM_MULTISHELL_PATH")
        if path then
            os.rmdir(path)
        end
    end
end)


-- doskey replacement
clink.onbeginedit(check_and_use_fnm)
```

bugficks:
I've updated the script: this [gist](https://gist.github.com/bugficks/0df37ebd3790e080ce6722a329245669) includes working doskey replacement and functions correctly even when launching the terminal using “Open Prompt Here” from Windows Explorer.

Schniz:
excuse me but what is clink

orzgithub:
> excuse me but what is clink

Here is it's GitHub repo: [chrisant996/clink](https://github.com/chrisant996/clink)
Clink is a tool for cmd.exe on Windows.
It brings tab completion functionality similar to bash, zsh, etc. into cmd, and also useful features such as color prompts for input commands. This tool also provides a function that can set some scripts written in Lua to be executed at startup, thereby achieving a startup script capability that is much more flexible than modifying the registry. It is directly injected into the cmd process, which means that it does not rely on specific terminal software like cmder. It can be used with any terminal like Windows Terminal, Windows legacy terminal, wezterm, Fluent Terminal, terminal components of various IDEs, etc. (cmder can of course also be used).

PR #1408, opened 2025-04-02
feat(env): prevent duplicated PATH from fnm env

fabricionaweb:
As previous discussed on #1309 (and maybe somewhere else), the current behavior is to always make a new symlink and append this new folder to PATH.

While I understand this works fine and one usually does not call `fnm env` multiple times for a shell session, I believe this has room to improve...

I made this change to be less aggressive. I didn't touch other variables and just attacking these two variables that are mutable and related.

On this PR I propose:

- `fnm env` will look for `fnm_multishells` on the environment variable `$PATH`.
If found it will not print the `PATH` variable, thus preventing it to be duplicated.

It also uses the value for set `FNM_MULTISHELL_PATH` but stripes out the `/bin`


fabricionaweb:
@Schniz does the latest release invalidate this? and the mentioned PR?

PR #1411, opened 2025-04-09
Fix fnm setup command in PowerShell profile

nicolasbalao:
While setting up the environment on Windows, I noticed that the command provided for adding fnm to the PowerShell profile was incorrect. fnm doesn't recognize --shell powershell — the correct value is --shell power-shell.

This PR updates the documentation to fix the command and ensure proper configuration for Windows users.


bjohansebas (approved):
I confirm that this command makes fnm work on Windows.

Issue #1376, opened 2025-02-04
Installing fnm on macOS using recommended script & Brew sets incorrect path


Mathias-S:
The fnm installation doesn't work on macOS when using the recommended script to install fnm (which uses Homebrew):

```
curl -fsSL https://fnm.vercel.app/install | bash
```

The script installs fnm to Homebrew's folder, typically `/opt/homebrew/bin/fnm` on macOS running ARM.

However, the installation script is hard coded to set `$FNM_PATH` to `$HOME/Library/Application Support/fnm` (which doesn't exist), so the code inserted into the shell file will have the incorrect path:

```bash
# fnm
FNM_PATH="/Users/myusername/Library/Application Support/fnm"
if [ -d "$FNM_PATH" ]; then
  export PATH="/Users/myusername/Library/Application Support/fnm:$PATH"
  eval "`fnm env`"
fi
```

This renders the installation broken unless the user manually corrects the path.

The [README also incorrectly states that fnm is installed in that folder](https://github.com/Schniz/fnm?tab=readme-ov-file#parameters).

transfer1992:
I have exactly this experience on macOS right now.

joseph-navant:
I just manually added `/opt/homebrew/bin/fnm env --use-on-cd | source` to my `fnm.fish` file

Tedfulk:
@joseph-navant This worked! Thank You!

joseph-navant:
> [@joseph-navant](https://github.com/joseph-navant) This worked! Thank You!

I finally moved this to the end of my `config.fish` because the previous workaround was giving me some issues when opening new tabs/windows:
```
# command hooks
fnm env --use-on-cd | source
```

babhishek21:
I fixed this by setting the `FNM_PATH` to `/Users/<my username>/.local/share/fnm` in the shell script that the install script installs. You can verify this would work, by first running `fnm env` (which is what is being `eval`ed / sourced) and checking the value of `FNM_DIR` in the output.

For my setup (fish shell), the sourcing happens in the `/Users/<my username>/.config/fish/conf.d/fnm` file:
```
# fnm
set FNM_PATH "/Users/<my username>/.local/share/fnm"
if [ -d "$FNM_PATH" ]
  set PATH "$FNM_PATH" $PATH
  fnm env | source
end
```

SeasideLee:
[...]

Issue #1279, opened 2024-10-10
Zsh shell setup command did not work for me


aleksebastian:
Installed via homebrew and tried adding `eval "$(fnm env --use-on-cd --shell zsh)"` to my .zshrc file per the documentation but running node -v on a new terminal would give me:

> zsh: command not found: node

Switching the provided command to the following that I found on my other computer finally made it work for me:
```
export PATH="/Users/work/Library/Application Support/fnm:$PATH"
eval "`fnm env`"
```

On an M3 MacBook Pro running macOS 15.0.1

AllySummers:
I experienced the same issue. For some reason this directory isn't getting created on initial install with homebrew?

Issue #1237, opened 2024-08-26
Install script configures .zshrc incorrectly on MacOS when using Homebrew


esseb:
Using macOS 14.6.1 (23G93)

Steps to reproduce:
1. Install using `curl -fsSL https://fnm.vercel.app/install | bash`
2. Run `source /Users/esseb/.zshrc` as instructed
3. Run `fnm use`

Expected result:
`fnm` should run without any errors.

Actual result:
```
error: We can't find the necessary environment variables to replace the Node version.
You should setup your shell profile to evaluate `fnm env`, see https://github.com/Schniz/fnm#shell-setup on how to do this
Check out our documentation for more information: https://fnm.vercel.app
```

Further information:
I think the bug is because `setup_shell` in https://fnm.vercel.app/install appends this to `.zshrc`:
```
# fnm
FNM_PATH="/Users/esseb/Library/Application Support/fnm"
if [ -d "$FNM_PATH" ]; then
  export PATH="/Users/esseb/Library/Application Support/fnm:$PATH"
  eval "`fnm env`"
fi
```

The folder that `FNM_PATH` refers to is never created in `download_fnm` when using homebrew, however, so the if statement is not true and `fnm env` does not get run.

esseb:
https://github.com/Schniz/fnm/issues/824#issuecomment-2291882222 seems to have discovered the same issue with the incorrect `FNM_PATH`.

TrevorBurnham:
Same issue here. Fresh install under zsh on macOS 15.1.1.

Can confirm that fix you linked to (changing the path in `~/.zshrc`) works. 👍🏻

Schniz:
Duplicate of #1194. Closing in favor of that issue.

---

*This message was written by AI. Sorry it's not personal, but it's helping me maintain this project. Feel free to disagree with the statement and reopen the issue if needed.*

Issue #1194, opened 2024-07-15
Fresh install on macOS using Homebrew: shell looking in wrong INSTALL_DIR


Scotchester:
I apologize if I'm missing something obvious.

I just tried installing fnm on a fresh MacBook (running macOS Sonoma 14.5) using the default recommended installation script (`curl -fsSL https://fnm.vercel.app/install | bash`). I can see that this uses Homebrew to do the installation. The output looks pretty expected:

```sh
cranfill@MT-407147 ~ % curl -fsSL https://fnm.vercel.app/install | bash
Downloading fnm using Homebrew...
Checking dependencies for the installation script...
Checking availability of curl... OK!
Checking availability of unzip... OK!
Checking availability of Homebrew (brew)... OK!
==> Downloading https://formulae.brew.sh/api/formula.jws.json
#=#=#                                                                           
==> Downloading https://formulae.brew.sh/api/cask.jws.json
######################################################################### 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/fnm/manifests/1.37.1
Already downloaded: /Users/cranfill/Library/Caches/Homebrew/downloads/4ff5bc74da6dcb0f0d50fd4ab8d966015c9c47dce2cca73cfed4709cb35e5035--fnm-1.37.1.bottle_manifest.json
==> Fetching fnm
==> Downloading https://ghcr.io/v2/homebrew/core/fnm/blobs/sha256:b76925bd68d738
Already downloaded: /Users/cranfill/Library/Caches/Homebrew/downloads/c3a21558106709d5104b0af89a1cdb099d9bb40af6d236169bb755eddd8382f3--fnm--1.37.1.arm64_sonoma.bottle.tar.gz
==> Pouring fnm--1.37.1.arm64_sonoma.bottle.tar.gz
==> Caveats
zsh completions have been installed to:
  /opt/homebrew/share/zsh/site-functions
==> Summary
🍺  /opt/homebrew/Cellar/fnm/1.37.1: 12 files, 7.2MB
==> Running `brew cleanup fnm`...
Disable this behaviour by setting HOMEBREW_NO_INSTALL_CLEANUP.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
Installing for Zsh. Appending the following to /Users/cranfill/.zshrc:

# fnm
FNM_PATH="/Users/cranfill/Library/Application Support/fnm"
if [ -d "$FNM_PATH" ]; then
[...]

PR #1414, opened 2025-04-24
feat(install): fix setup_shell function for brew users

### LINKED ISSUE #824: `node: command not found` on MacOS 12
Whether it's [`manual`](https://github.com/Schniz/fnm#manually) installation or [`fnm-install.sh`](https://github.com/Schniz/fnm/blob/master/.ci/install.sh) installation or [`fnm-install.sh`](https://github.com/Schniz/fnm/blob/master/.ci/install.sh) installation, and [shell-setup](https://github.com/Schniz/fnm#shell-setup).

I found sample problem on [#427](https://github.com/Schniz/fnm/issues/427) , but no solution found. 

In addition, the directory of `$FNM_MULTISHELL_PATH/bin`  also doesn't exist.

System is `macos 12`

better-salmon:
# Fixing the install script for brew users

## Issue
When installing `fnm` via Homebrew, the binary gets installed to Homebrew's directory structure rather than the default `INSTALL_DIR`. Additionally, Homebrew automatically adds `fnm` to the `PATH`, making our manual `PATH` manipulation unnecessary and potentially problematic.

## Changes
- Set a `USE_HOMEBREW` flag to false by default
- Updated shell configuration for all supported shells (zsh, fish, bash) to handle Homebrew installations differently:
  - For Homebrew installations: only add `eval "$(fnm env)"` to shell config files
  - For direct installations: maintain existing behavior with `PATH` manipulation
- Simplify shell configuration when Homebrew is used, avoiding redundant `PATH` modifications

These changes ensure `fnm` works correctly regardless of installation method, while respecting the environment setup performed by package managers.

Fix #824 #1190 #1194 #1237 #1279 #1376

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -4,6 +4,7 @@ set -e
 
 RELEASE="latest"
 OS="$(uname -s)"
+USE_HOMEBREW="false"
 
 case "${OS}" in
    MINGW* | Win*) OS="Windows" ;;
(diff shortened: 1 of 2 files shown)


polybuildr:
👋 @Schniz looks like this PR and https://github.com/Schniz/fnm/pull/1362 both attempt to fix this issue, do you think one of them could be merged soon? Fresh installs following the instructions at https://nodejs.org/en/download still seem to fail on MacOS due to these bugs.

.ci/install.sh
```diff
@@ -4,6 +4,7 @@ set -e
 
 RELEASE="latest"
 OS="$(uname -s)"
+USE_HOMEBREW="false"
```
Schniz, line 7: why shouldn't homebrew be the default though?

PR #1421, opened 2025-05-07
Fix `powershell` typo in README.md

Ankuriboh:
On fnm, shell possible values are: bash, zsh, fish, power-shell, cmd.

This PR update README.md to reflect this.

PR #1434, opened 2025-06-04
fix(shell): actually allow `powershell` and `power-shell` as shell type inputs

### LINKED ISSUE #1040: Completions: invalid value `powershell` for `--shell` option
Hello.

I recently updated from 1.33.1 to 1.35.1, and somewhere in between seems to have changed from `powershell` to `power-shell`.
I'm not sure if it was an intended change but in `README.md` it still shows `powershell` as the correct option.

image

Kind regards!

### LINKED ISSUE #1346: The setup script is different for PowerShell7
Suggest corrections to the documentation.

https://github.com/Schniz/fnm?tab=readme-ov-file#powershell

For PowerShell7, we need to specify `power-shell` (not `powershell`) with `--shell`. 😫😫😫

### LINKED ISSUE #1431: Wrong config instruction for powershell setup in readme
Current line in the readme, to add to the shell configurations:

```
fnm env --use-on-cd --shell powershell | Out-String | Invoke-Expression
```

This is the right line to add to the shell

```
fnm env --use-on-cd --shell power-shell | Out-String | Invoke-Expression
```

SunsetTechuila:
fixes #1431
fixes #1346
fixes #1040

related to #1264

supersedes #1411, #1421, #1423

### DIFF
--- a/.changeset/crazy-rivers-rest.md
+++ b/.changeset/crazy-rivers-rest.md
@@ -1,0 +1,5 @@
+---
+"fnm": patch
+---
+
+actually allow both `powershell` and `power-shell` as shell type inputs
--- a/src/shell/shell.rs
+++ b/src/shell/shell.rs
@@ -18,7 +18,7 @@ pub enum Shells {
     Bash,
     Zsh,
     Fish,
-    #[clap(name = "powershell", alias = "power-shell")]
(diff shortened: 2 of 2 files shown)


SunsetTechuila:
Just noticed that I had an outdated version (1.35.1). I installed fnm from the winget, like probably other people who keep reporting this problem

PR #1446, opened 2025-07-02
ci(command-docs): fix path resolution on windows

SunsetTechuila:


SunsetTechuila:
there's another issue: on windows, the script adds `cmd` to the shell list, but the `Ensure command docs are up-to-date` job runs on linux, so the check fails

Issue #1469, opened 2025-10-21
`--corepack-enabled` doesn't work with Node.js 25


jessestricker:
Installing Node.js 25 (or later) does not work when also using the `--corepack-enabled` flag.
This is because as of Node.js 25, corepack is no longer bundled in the Node.js installation and has to be installed separately.

To reproduce, run the following:

```console
$ fnm --version
fnm 1.38.1

$ fnm install 25 --corepack-enabled
Installing Node v25.0.0 (x64)
Enabling corepack for Node v25.0.0
error: Can't enable corepack: Can't spawn program: No such file or directory (os error 2)
Maybe the program /home/jesse/.local/share/fnm/node-versions/v25.0.0/installation/bin/corepack does not exist on not available in PATH?
```

kazaz-lumida:
Any chance this will get some attention now that 26.x.x is out? :)

Issue #139, opened 2019-09-09
[Request] Automatically install default packages


chocolateboy:
(There are some [similar](https://github.com/Schniz/fnm/issues/109)-sounding [issues](https://github.com/Schniz/fnm/issues/80), but they're talking about *sharing* global packages across different versions rather than installing them per-version.)

It'd be nice to be able to automatically (re-)install a list of global packages when a new version is installed as can be done via the [default-packages](https://github.com/nvm-sh/nvm/blob/v0.34.0/README.md#default-global-packages-from-file-while-installing#default-global-packages-from-file-while-installing) file in nvm:

> If you have a list of default packages you want installed every time you install a new version we support that too. You can add anything npm would accept as a package argument on the command line.

e.g.:

```bash
$ cat $NVM_DIR/default-packages
```
```
jay-repl
tmpin
web-ext
```

Or, failing that, some kind of post-install hook?

Schniz:
As I don't use global packages, I don't have this kind of problems. I think that using global packages that are linked to the current Node version is mostly a user configuration smell. It is exactly like global variables. Most packages should be installed as a devDependency or a direct dependency to the project you're working on. It's easier to manage, and if you need an escape hatch, we have tools like `npx`.

Anyway, thanks for opening the issue. It keeps showing me that my way of using the computing machine is not the only way to do so, and things need to get better. However, I don't think we should allow bad practices.

> It'd be nice to be able to automatically (re-)install a list of global packages when a new version is installed as can be done via the [default-packages](https://github.com/nvm-sh/nvm/blob/v0.34.0/README.md#default-global-packages-from-file-while-installing) file in nvm

This is an interesting idea but would make installation much, much slower. Also, it will produce inconsistencies and I'm not sure this will be used eventually, because you can have multiple versions (node 6 will have `tsc@2`, node 8 will have `tsc@3.5`, node 10 will have `tsc@3.6` etc).

[This comment](https://github.com/Schniz/fnm/issues/109#issuecomment-499088976) suggests a different solution: having a new binary that will work like `gnpm install tmpin` will install `tmpin` using the _current node version_, and add a small bash script that fixes _the current node version_ into `$FNM_DIR/global-packages/tmpin` or any binary listed in `package.json#bin`. Maybe `gnpm` isn't a good name. Maybe `fnmpm`? idk. I'm bad with naming!

[...]

PR #1503, opened 2026-02-12
feat: add default-packages file support for fnm install

### LINKED ISSUE #1093: Feature request: Port over nvm's "default global packages"
`nvm` has a feature that I would find useful in my current developer setup. The specific feature is [default global packages](https://github.com/nvm-sh/nvm#default-global-packages-from-file-while-installing) that are installed on every new version of node setup.

Our development environment at my current job relies on a few globally-installed npm packages and it can be a headache to remember to install them when jumping to a new node version, or jumping to an older version that has been previously cleared.

amanthanvi:
## Summary

Adds support for a `default-packages` file (`$FNM_DIR/default-packages`) that is
read after every `fnm install`. Packages listed in this file are automatically
installed globally via `npm install -g` in the context of the newly installed
Node version.

This is the most requested missing feature in fnm, tracked since 2019.

## Motivation

- Closes #1093
- Addresses #139 (the original 2019 request)
- Compatible with nvm's `$NVM_DIR/default-packages` format
- Users switching from nvm can reuse their existing default-packages file

## Implementation

Follows the exact pattern established by `enable_corepack()` in `install.rs`:
- Reads the file from `config.default_packages_file()` (new method on `FnmConfig`)
- Parses lines (skipping comments and blanks)
- Runs `npm install -g <packages>` via `Exec::new_for_version()`
- Placed after corepack enablement, before `--use` activation
- Silently no-ops if the file is missing (opt-in feature)

## File format

```
# Comments supported
typescript
prettier@3
@openai/codex
eslint
```

## Testing

- Unit tests for the parser (comments, blanks, scoped packages, @version)
- E2E test modeled on corepack.test.ts (installs a test package, verifies presence)
- Cross-platform: Unix (npm) + Windows (npm.cmd) paths handled

## Checklist

- [ ] `cargo test` passes
- [ ] `cargo clippy --all-targets` clean
- [ ] `cargo fmt --check` clean
- [ ] E2E tests pass
- [ ] Docs updated (configuration.md, commands.md)
- [ ] Changeset added


jessestricker:
This would also fix the --enable-corepack flag for Node v25 and newer, by being able to automatically install corepack for these versions which do not bundle it.

Closes #1469

indykoning:
> This would also fix the --enable-corepack flag for Node v25 and newer, by being able to automatically install corepack for these versions which do not bundle it.
> 
> Closes #1469

In it's current implementation it couldn't, packages are attempted to be installed **after** attempting to enable Corepack. So you will get an error before it has the chance to install the global packages.
I do agree that it would be a nice addition though, perhaps just adding it to the global packages if the config of enable_corepack is true

Issue #1094, opened 2024-01-26
Change directory not triggering use of default node version


yusuphgammut:
Hi fnm team. First congrats for writing and maintaining this tool. It's supper useful and fast.

I've just installed it (adding `eval "$(fnm env --use-on-cd)"` to my `~/.zshrc` file) and I noticed that when I enter a project that has a node version specified via `.nvmrc` file it automatically triggers the use of that version, but when I move outside of that folder it doesn't trigger the use of the default one. This is what I see:

``` sh
> fnm list
* v18.10.0
* v20.11.0 default
* system
> node --version
v20.11.0
> cd my-node-project
Using Node v18.10.0
> node --version
v18.10.0
> cd ..
> node --version
v18.10.0
```

I'm not sure if this is the desired behavior if it's an issue. I used nvm till today and it triggered the use of the default node version when moving outside the project.

Additional info:

- OS: macOS v14.3
- Terminal: Alacritty v0.13.0-dev
- Shell: Zsh v5.9
- FNM: v1.35.1 installed via Homebrew v4.2.5

gpetrioli:
I am having the same issue.

However, as a workaround, you can the `--version-file-strategy` to  `recursive` and then go to your top folder (_i went to `~`_) and add a `.nvmrc` file there.

To set the flag, edit your `.nvmrc` file and edit the 

```
eval "$(fnm env  --use-on-cd)"
```
to 
```
eval "$(fnm env --version-file-strategy recursive --use-on-cd)"
```

nathanpovo:
@Schniz I have created a PR to update the documentation to make it easier for users to set up their shell so that they do not run into this issue.

Sorry for just creating a PR without asking but I thought I would just do it quickly instead of asking and then having to waste time context switching back to this issue.

Feel free to discard the PR if you do not agree with the changes.

Issue #680, opened 2022-02-23
Document or set the recursive strategy as default with --use-on-cd


NicolasWebDev:
I had been waiting for https://github.com/Schniz/fnm/pull/607 to be merged, and it is actually working really well.

But I think that for new people, --use-on-cd should use the recursive strategy by default, because I assume that it is the expected behavior to go back to default when leaving the directory. Maybe it could induce a breaking change though.

In any case, mentioning it in the documentation would be quite helpful for newcomers I think.

Thanks to that feature, I can finally ditch nvm and get back 500ms of startup time on my shell, so thanks for the great work!

taylorkline:
@Schniz Would a PR to update the `README.md` to mention that the `--version-file-strategy=recursive` flag may be desired with the `--use-on-cd` addition to the shell configs?

PR #1504, opened 2026-02-12
fix: --use-on-cd reverts to default version when leaving versioned directories

amanthanvi:
## Summary

Adds an `else` branch to the `--use-on-cd` shell hooks for the `local` version-file
strategy. When entering a directory without `.node-version` or `.nvmrc`, fnm now reverts
to the default Node version instead of staying on the last project version.

## Motivation

This is the most repeatedly reported UX issue in fnm:
- #442 — @Schniz said "I can add the default version if it finds nothing"
- #320 — "use-on-cd doesn't switch back to default"
- #144 — "Better support for automatic switching"
- #1094 — "Change directory not triggering use of default node version"
- #680 — Requests making this the default behavior

The `recursive` strategy already handles this (falls back to `default_version` in
`version_files.rs`), but the `local` strategy (the default, and the most common setup)
did not. Users expect auto-switching to be bidirectional: switch when entering a versioned
directory, revert when leaving.

## Implementation

Each shell hook (bash, zsh, fish, powershell) for the `local` strategy previously had:

```bash
if [[ -f .node-version || -f .nvmrc ]]; then
    fnm use --silent-if-unchanged
fi
```

Now (bash shown; other shells use equivalent syntax):

```bash
if [[ -f .node-version || -f .nvmrc ]]; then
    fnm use --silent-if-unchanged
else
    if [[ -e "$FNM_DIR/aliases/default" ]]; then
        fnm use "$(fnm default)" --silent-if-unchanged
    fi
fi
```

This keeps the `local` strategy semantics (only checks the current directory), while
still reverting to the configured default version when no version file is present.

## Edge cases handled

| Scenario | Behavior |
|---|---|
| cd to dir WITH version file | Switches to that version (unchanged) |
| cd to dir WITHOUT version file | Reverts to default version (NEW) |
| Already on default, cd to non-versioned dir | Silent no-op |
| No default version configured | Silent no-op |
| User manually ran `fnm use X`, then cd to non-versioned dir | Reverts to default (matches nvm) |

## Note on "manual fnm use" edge case

In #320, concern was raised about reverting when a user manually ran `fnm use X` and
then cd'd to a subdirectory. This PR does revert in that case, matching nvm's behavior.
The rationale: `--use-on-cd` means "manage my version based on directory." If users want
to pin a version regardless of directory, they should use `fnm use X` without
`--use-on-cd`, or add a `.node-version` file to that directory.

## Testing

- E2E: revert to default when leaving versioned dir (bash, zsh, fish, powershell)
- E2E: no error when no default configured
- E2E: silent when already on default
- Unit: updated hook string assertions (if existing tests cover hook generation)
- `cargo test` + `cargo clippy` + `cargo fmt` all clean

## Checklist

- [ ] `cargo test` passes
- [ ] `cargo clippy --all-targets` clean
- [ ] `cargo fmt --check` clean
- [ ] E2E tests pass across shells
- [ ] All four shell hooks updated (bash, zsh, fish, powershell)
- [ ] Docs updated
- [ ] Changeset added

### DIFF
--- a/.changeset/sticky-lemurs-revert.md
+++ b/.changeset/sticky-lemurs-revert.md
@@ -1,0 +1,6 @@
+---
+"fnm": patch
+---
+
+Fixed `--use-on-cd` with `local` strategy to revert to the default Node version when entering a directory without a version file. Previously, the Node version would remain "stuck" on the last project version after leaving a versioned directory. This now matches nvm's auto-switching behavior.
+
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -414,6 +414,10 @@ Options:
       --use-on-cd
           Print the script to change Node versions every directory change
 
+          When entering a directory with a version file, fnm switches to that version. When entering a directory without a version file, fnm switches to the default version.
+
+          This applies to both `local` and `recursive` version file strategies.
+
       --arch <ARCH>
           Override the architecture of the installed Node binary. Defaults to arch of fnm binary
 
--- a/e2e/use-on-cd-revert-default.test.ts
+++ b/e2e/use-on-cd-revert-default.test.ts
@@ -62,12 +62,12 @@ for (const shell of [Bash, Zsh, Fish, PowerShell]) {
       const captureCdOutput = (() => {
         const outputFile = "cd-output.txt"
         if (shell === Fish) {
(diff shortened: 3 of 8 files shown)


mikibakaiki:
Is this going to be continued? :)

PR #1516, opened 2026-03-05
feat(env): apply initial use during env --use-on-cd

Schniz:
## Why
`fnm env --use-on-cd` previously relied on shell hook output to immediately spawn an extra `fnm use` subprocess after eval. This change applies the initial version inside `fnm env` itself, reducing shell startup overhead and making initialization smoother.


fabricionaweb:
This seems to close #1408 and #1309 if I get it correct, since it have been a year Im not sure :P

PR #1523, opened 2026-03-09
Clarify interactive install prompt source

### LINKED ISSUE #1521: Add hint to change directory hook
After I cd to some repository dir and I saw this message.

Image

My first thought was: "What the heck is this? Who's saying? Where will it be installed?"

So what about adding hint like:

```
fnm can't find an installed Node version matching v24.11.1.
Do you want to install it? answer [y/N]: y
```

Schniz:
## Summary
- prefix the missing-version interactive message with `fnm` so it's clear who is prompting
- keep the existing install confirmation flow unchanged while improving first-time readability

Closes #1521

### DIFF
--- a/.changeset/fair-carrots-greet.md
+++ b/.changeset/fair-carrots-greet.md
@@ -1,0 +1,5 @@
+---
+"fnm": patch
+---
+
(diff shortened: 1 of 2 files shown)


devxoul:
Love this. Thanks!!

PR #1535, opened 2026-03-21
PowerShell `--use-on-cd` enhancements

### LINKED ISSUE #1350: Use existing `cd` alias in the script that `fnm env --shell powershell --use-on-cd` emits
## Problem

On PowerShell, `fnm env` currently emits the following 👇, which alises `cd` to `Set-LocationWithFnm`:
```powershell
# environment variables set with $env:... = ...
function global:Set-FnmOnLoad { fnm use --silent-if-unchanged }
function global:Set-LocationWithFnm { param($path); if ($path -eq $null) {Set-Location} else {Set-Location $path}; Set-FnmOnLoad }
Set-Alias -Scope global cd_with_fnm Set-LocationWithFnm
Set-Alias -Option AllScope -Scope global cd Set-LocationWithFnm
Set-FnmOnLoad
```

Consequently, if `cd` is already aliased to a different command (e.g., as part of one's dotfiles, or a different command that similarly "patches" `cd`), `fnm env | iex` drops the current alias and kinda "resets" it back to `Set-Location`.

For example,

1. `cd` is aliased to print out the path being navigated to, as a contrived example:
   ```powershell
   function global:customCd {
   	param ($path)
   	Write-Host "customCd: $path"
   	Set-Location $path
   }
   Set-Alias 'cd' 'customCd' -Option AllScope -Scope Global
   ```
   
   ...which has the following effect:
   ```
   PS ~> cd Projects
   customCd: Projects
   PS ~/Projects>
   ```

2. FNM is loaded with `fnm env --shell powershell --use-on-cd | Out-String | Invoke-Expression`.

3. `cd` now prints:
   ```
   PS ~> cd Projects/jsProject
   Using Node v22.12.0
   PS ~/Projects/jsProject>
   ```
   ...and `global:customCd` from step 1 is no longer invoked.

## Proposed Solution

Thoughts on emitting a script like 👇? I've only included the `cd`-related bits; `Set-FnmOnLoad` and `Set-LocationWithFnm` are unchanged and omitted for brevity.

```powershell
$global:cdBeforeFnm = (Get-Alias 'cd' -Scope Global -ErrorAction Ignore).ReferencedCommand ?? (Get-Command Set-Location)

function global:cdWithFnm {
	param ($path)
	if ($null -eq $path) { & $global:cdBeforeFnm } else { & $global:cdBeforeFnm $path }
	global:Set-FnmOnLoad
}

Set-Alias 'cd' global:cdWithFnm -Option AllScope -Scope Global
```

...which uses the command that `cd` references at the time FNM is loaded, instead of always using `Set-Location`, also invoking `global:customCd`:

```
PS ~> cd Projects/jsProject
customCd: Projects/jsProject
Using Node v22.12.0
PS ~/Projects/jsProject>
```

ravindUwU:
This PR includes the following changes for the `fnm env --use-on-cd` script emitted for PowerShell:
- The `cd` command applicable when the `fnm env` script is sourced is invoked by the emitted function, instead of `Set-Location` directly. Fixes #1350.
- All params of the emitted `cd` function are forwarded to the underlying `cd` function using [splatting](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.5#splatting-command-parameters).

## Demo

With the following PowerShell profile,

```powershell
# customCd prints its args to cd before continuing with Set-Location.
function customCd { Write-Host "customCd: [$args]" ; Set-Location @args }
Set-Alias -Option AllScope -Scope Global 'cd' 'customCd'

fnm env --shell powershell --use-on-cd | Out-String | Invoke-Expression
```

Notice how, on PowerShell 5 & 7,
- `cd`-ing uses FNM on cd, and delegates to the `customCd` function.
- Parameters of the underlying `cd` function (in this case the `-PassThru` option of `Set-Location`) can be specified.

### DIFF
--- a/src/shell/powershell.rs
+++ b/src/shell/powershell.rs
@@ -42,10 +42,12 @@ impl Shell for PowerShell {
         };
         Ok(formatdoc!(
             r"
-                $global:cd_before_fnm = (Get-Command 'cd' -ErrorAction Ignore)
-                if (-not $global:cd_before_fnm) {{ $global:cd_before_fnm = 'Set-Location' }}
                 function global:Set-FnmOnLoad {{ {autoload_hook} }}
-                function global:Set-LocationWithFnm {{ & $global:cd_before_fnm @args; Set-FnmOnLoad }}
+                & {{
+                    $cd = (Get-Command 'cd' -ErrorAction Ignore)
+                    if (-not $cd) {{ $cd = 'Set-Location' }}
+                    Set-Item Function:\global:Set-LocationWithFnm {{ & $cd @args; Set-FnmOnLoad }}.GetNewClosure()
+                }}
                 Set-Alias -Scope global cd_with_fnm Set-LocationWithFnm
                 Set-Alias -Option AllScope -Scope global cd Set-LocationWithFnm
             ",
--- a/src/shell/powershell.rs
+++ b/src/shell/powershell.rs
@@ -45,7 +45,7 @@ impl Shell for PowerShell {
                 $global:cd_before_fnm = (Get-Command 'cd' -ErrorAction Ignore)
                 if (-not $global:cd_before_fnm) {{ $global:cd_before_fnm = 'Set-Location' }}
                 function global:Set-FnmOnLoad {{ {autoload_hook} }}
-                function global:Set-LocationWithFnm {{ param($path); if ($path -eq $null) {{ & $global:cd_before_fnm }} else {{ & $global:cd_before_fnm $path }}; Set-FnmOnLoad }}
+                function global:Set-LocationWithFnm {{ & $global:cd_before_fnm @args; Set-FnmOnLoad }}
                 Set-Alias -Scope global cd_with_fnm Set-LocationWithFnm
                 Set-Alias -Option AllScope -Scope global cd Set-LocationWithFnm
             ",
--- a/src/shell/powershell.rs
+++ b/src/shell/powershell.rs
@@ -42,8 +42,10 @@ impl Shell for PowerShell {
         };
         Ok(formatdoc!(
             r"
+                $global:cd_before_fnm = (Get-Command 'cd' -ErrorAction Ignore)
+                if (-not $global:cd_before_fnm) {{ $global:cd_before_fnm = 'Set-Location' }}
                 function global:Set-FnmOnLoad {{ {autoload_hook} }}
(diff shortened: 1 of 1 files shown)


ravindUwU:
What's emitted now is 😄:

```pwsh
function global:Set-FnmOnLoad { ... }
& {
	$cd = (Get-Command 'cd' -ErrorAction Ignore)
	if (-not $cd) { $cd = 'Set-Location' }
	Set-Item Function:\global:Set-LocationWithFnm { & $cd @args; Set-FnmOnLoad }.GetNewClosure()
}
Set-Alias -Scope global cd_with_fnm Set-LocationWithFnm
Set-Alias -Option AllScope -Scope global cd Set-LocationWithFnm
```

The `cd` command before FNM is captured as `$cd` within the `& { ... }` scope, and `Set-LocationWithFnm` is made within and bound to that scope. Thoughts, @HeyItsGilbert? 😅

Re: `cd_with_fnm`, thoughts on removing this alias? 🤔 It _could_ be useful if a different tool "resets" `cd` (as FNM did before); but that behaviour should be considered a bug with the other tool and fixed similarly, IMO.

HeyItsGilbert:
Looks good to me. I agree that you could remove the alias.

Realistically what I'm going to probably end up doing is figuring out what zoxide is doing and merging that and the Set-FnmOnLoad command into my custom one.

ravindUwU:
Alternatively, I think `Set-LocationWithFnm` could be removed, and `cd_with_fnm` made a function; like,

```powershell
function global:Set-FnmOnLoad { ... }
& {
	$cd = (Get-Command 'cd' -ErrorAction Ignore)
	if (-not $cd) { $cd = 'Set-Location' }
	Set-Item Function:\global:cd_with_fnm { & $cd @args; Set-FnmOnLoad }.GetNewClosure()
}
Set-Alias -Option AllScope -Scope global cd cd_with_fnm
```

...in fact I think I prefer this! The name `cd_with_fnm` clarifies that the underlying command is `cd`; whereas `Set-LocationWithFnm` is now a bit misleading in that it actually invokes `cd` instead of `Set-Location` 🤔.

A search for [`lang:PowerShell /Set-LocationWithFnm|cd_with_fnm/`](https://github.com/search?q=lang%3APowerShell%20%2FSet-LocationWithFnm%7Ccd_with_fnm%2F&type=code) shows a few scripts that reference both of these so there _might_ be some breakage if either is removed.

ravindUwU:
Hi @Schniz, interested in your thoughts on this PR, and the suggestion at https://github.com/Schniz/fnm/pull/1535#issuecomment-4146077395 😅.

Issue #1554, opened 2026-05-11
[BUG] Corepack Error on Win11 with Powershell config. C:\Users\username\AppData\Roaming\fnm\node-versions\v25.9.0\installation\corepack.cmd"' is not recognized as an internal or external command, operable program or batch file.


futuremotiondev:
I installed fnm via downloading the binary and added to PATH.

Added the following to my Pwsh profile:

```
fnm env --use-on-cd --version-file-strategy=recursive --corepack-enabled --shell powershell | Out-String | Invoke-Expression
```

Went to install v25.9.0 with `fnm install 25.9.0`

Node installed fine, but then in terminal:

```
Enabling corepack for Node v25.9.0
'"C:\Users\username\AppData\Roaming\fnm\node-versions\v25.9.0\installation\corepack.cmd"' is not recognized as an internal or external command, operable program or batch file.
```

I navigated to `C:\Users\username\AppData\Roaming\fnm\node-versions\v25.9.0\installation\` and there is no `corepack.cmd`. How can I enable corepack?

You should probably look at why corepack.cmd isn't being created despite the `--corepack-enabled` flag being set.

Any help appreciated.

o-l-a-v:
`fnm install 25.9.0 --corepack-enabled --use`

Edit: Never mind. Corepack is not included since Node.js 25.

* <https://github.com/Schniz/fnm/issues/1469>

Issue #1398, opened 2025-03-19
Install `packageManager` (replace corepack)


styfle:
Since [corepack is being phased out starting in node@25](https://github.com/nodejs/TSC/pull/1697#issuecomment-2737093616), it would be great if fnm could manage both the node version and the package manager (for example pnpm) in order to replace corepack.

Ideally both `package.json#packageManager` as well as `package.json#devEngines.packageManger` should be read by fnm.

SukkaW:
> No, but it's no longer part of node, so it doesn't really make sense for a node version manager to interact with it or to install it.

If so, it also makes no sense for a node version manager to interact with package managers or install them as well. Neither pnpm nor yarn is a part of Node.js in the very first place.

ljharb:
Yes, that's also correct - but this isn't isn't about those.

rose-bouchayer:
I do understand that the current scope of this project is only to manage Node.js version but sadly I had to migrate from fnm to [proto](https://moonrepo.dev/proto) in order to handle all my JavaScript stack versions (Node.js + pnpm).

It would be a great addition to fnm to make it easily manage the whole stack, from engine to package manager.

o-l-a-v:
@zanminkian Can engines pin version to SHA, like packageManager can?

kristof-mattei:
> [@zanminkian](https://github.com/zanminkian) Can engines pin version to SHA, like packageManager can?

I've tried, and in my research there doesn't seem to be such a mechanism.

SukkaW:
> > [@zanminkian](https://github.com/zanminkian) Can engines pin version to SHA, like packageManager can?
> 
> I've tried, and in my research there doesn't seem to be such a mechanism.

Then another reason to stick with corepack.

karlhorky:
Maybe `devEngines.packageManager` is not the desired place for the integrity hash.

pnpm has [`devEngines.packageManager` support](https://pnpm.io/package_json#devenginespackagemanager) and allow an integrity hash in that field, but only strips it:

https://github.com/pnpm/pnpm/blob/fecaee0b3591c48f8662f07bffbab73e1671e69a/config/reader/src/index.ts#L803-L804

[...]

PR #1566, opened 2026-06-03
Skip enabling Corepack if it's not bundled with Node.js

moposx:
Previously fnm ran `corepack enable` directly, which could lead to a "No such file or directory" error when the installed Node version did not bundle Corepack.

Now the installed version is checked against the semver range `^14.19.0 || 16.9.0 - 24.x.x`. If the version doesn't bundle corepack, fnm will just skip enabling Corepack and leave the user a hint to install and enable Corepack manually via `npm install -g corepack && corepack enable`.

This will serve as a workaround as #1398, #1469 are still open, and can provide clarity if a user runs into issues like #1554.


indykoning:
With the `--corepack-enabled` flag explicitly stating i want Corepack to be enabled on all my projects, i'd expect as a user that it would be installed for me if it is not installed.
Especially for non-technical users and pipelines it would be nice to have that managed automatically.

codler:
@Schniz can you have a look?

Schniz:
Sorry; I will hopefully look tmrw. I think I’d want to install corepack if not baked into node. Wild times

PR #1579, opened 2026-07-22
Don't pass duplicate /d flag in Windows cmd use-on-cd hook

### LINKED ISSUE #1556: `cd /d D:\` doesn't work
Problem: The generated cd.cmd always prepends `/d`, which can conflict with user-supplied `/d` and break commands such as `cd /d <drive>`.

https://github.com/Schniz/fnm/blob/a53186d1df9b93487f28744e6ff54c7b7e6cd41e/src/shell/windows_cmd/cd.cmd#L2

Fix: Scan arguments first; if `/d` already exists, run `cd %*`; otherwise, run `cd /d %*`.

```cmd
@echo off
set "_use_d=1"
call :detect_d %*
set "_use_d_local=%_use_d%"
set "_use_d="

if "%_use_d_local%"=="1" goto _cd_with_d
cd %*
goto _after_cd
:_cd_with_d
cd /d %*
:_after_cd
set "_use_d_local="
if errorlevel 1 goto _end

if /i "%FNM_VERSION_FILE_STRATEGY%"=="recursive" goto _fnm_use
if exist .nvmrc goto _fnm_use
if exist .node-version goto _fnm_use
goto _end

:_fnm_use
fnm use --silent-if-unchanged

:_end
exit /b

:detect_d
if "%~1"=="" goto :eof
if /i "%~1"=="/d" set "_use_d=0"
shift
goto detect_d
```

emerson-d-lopes:
Fixes #1556

## Problem

The generated `cd.cmd` always runs `cd /d %*`. When the user supplies their own `/d` (e.g. `cd /d D:\`), the flag is passed twice and cmd fails with:

```
The syntax of the command is incorrect.
```

## Fix

Add `/d` only when the first argument isn't already `/d`:

```cmd
if /i "%~1" == "/d" cd %*
if /i not "%~1" == "/d" cd /d %*
```

The shape is deliberate — each alternative fails in a non-obvious way (all verified empirically on Windows 11 cmd.exe):

- **No parenthesized `if/else` blocks**: `%*` may contain `)` — `cd /d C:\Program Files (x86)` inside a `( ... )` block closes the block early and breaks the command. The current single-line form handles it; a block would regress it.
- **No `setlocal`**: `endlocal` reverts directory changes, which would break `cd` entirely.
- **No `goto`/labels**: the file is stored with LF endings and embedded byte-exact via `include_bytes!`; cmd's label scanning is unreliable with LF-only line endings. The two guarded ifs avoid labels altogether.
- **First-argument check only** (rather than scanning all args): native `cd` itself only accepts `/d` before the path (`cd C:\ /d` fails with "The system cannot find the path specified"), so positions after the first never work anyway and are passed through to native `cd` to produce its normal error.

## Testing

WinCmd e2e tests are currently skip-listed (`e2e/describe.ts`), so I verified the behavior directly on Windows 11 with a 15-case matrix against real `cmd.exe`, comparing old script vs new script vs native `cd`:

| Case | Old | New |
|---|---|---|
| `cd /d Y:\` (cross-drive via `subst`, dir contains `.node-version`) | ❌ syntax error | ✅ cd works + `fnm use --silent-if-unchanged` fires |
| `cd /D Y:\` (uppercase) | ❌ | ✅ |
| `cd ..`, `cd \`, `cd` (no args) | ✅ | ✅ unchanged |
| `cd "C:\Program Files"` / unquoted `cd C:\Program Files` | ✅ | ✅ unchanged |
| `cd C:\Program Files (x86)` and `cd /d C:\Program Files (x86)` | ✅ / ❌ | ✅ / ✅ |
| nonexistent dir | stays put + native error | unchanged |
| `.node-version` / `.nvmrc` / `FNM_VERSION_FILE_STRATEGY=recursive` triggers | ✅ | ✅ unchanged (verified with a stubbed `fnm` on PATH) |
| environment variable leakage into the session | none | none (fix uses no variables) |

Also verified:

- The LF-endings variant of the script (what release binaries embed) passes the same cases.
- `cargo test` — 28/28 pass, including `shell::windows_cmd::tests::use_on_cd_quotes_macro_path`.
- Built `fnm.exe env --use-on-cd --shell cmd` generates the fixed script; re-ran key cases against the generated file.
- Added `e2e/repros/issue-1556.test.ts` per AGENTS.md (WinCmd-only; it self-skips today but will cover this once WinCmd e2e is enabled). Compiles under jest, prettier clean.
- Changeset included (patch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01U5TQ1y3yBEDx73tYTHoUBj

### DIFF
--- a/.changeset/witty-drives-turn.md
+++ b/.changeset/witty-drives-turn.md
@@ -1,0 +1,10 @@
+---
+"fnm": patch
+---
+
+fix `cd /d <drive>` failing in Windows Command Prompt when use-on-cd is enabled: the generated `cd.cmd` always prepended `/d`, so a user-supplied `/d` was passed twice and `cd` failed with "The syntax of the command is incorrect". Fixes #1556
+
+```sh-session
+> cd /d D:\
+The syntax of the command is incorrect.
+```
--- a/e2e/repros/issue-1556.test.ts
+++ b/e2e/repros/issue-1556.test.ts
@@ -1,0 +1,22 @@
+import { writeFile, mkdir } from "node:fs/promises"
+import { join } from "node:path"
+import { script } from "../shellcode/script.js"
+import { WinCmd } from "../shellcode/shells.js"
+import testCwd from "../shellcode/test-cwd.js"
+import testNodeVersion from "../shellcode/test-node-version.js"
+import describe from "../describe.js"
+
+describe(WinCmd, () => {
+  test(`issue #1556: cd /d works with use-on-cd`, async () => {
+    const subdir = join(testCwd(), "subdir")
+    await mkdir(subdir, { recursive: true })
+    await writeFile(join(subdir, ".node-version"), "v12.22.12")
+    await script(WinCmd)
+      .then(WinCmd.env({ useOnCd: true }))
+      .then(WinCmd.call("fnm", ["install", "v8.11.3"]))
+      .then(WinCmd.call("fnm", ["install", "v12.22.12"]))
+      .then(WinCmd.call("cd", ["/d", subdir]))
+      .then(testNodeVersion(WinCmd, "v12.22.12"))
+      .execute(WinCmd)
+  })
+})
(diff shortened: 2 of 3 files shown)


jasonnicholson:
I support this change. Thank you.

PR #1595, opened 2026-09-12
fix: use POSIX compliant [ ] syntax instead of bash-only [[ ]] in env --use-on-cd

### LINKED ISSUE #1591: `fnm env --use-on-cd` emits bash-only `[[ ]]` syntax with no `POSIX` sh option
**Environment**: fnm 1.30.0, Ubuntu 24.04, dash (/bin/sh)

In general, any Linux environment where /bin/sh -> dash

**Bug**

`fnm env --use-on-cd` generates:

```
if [[ -f .node-version || -f .nvmrc || -f package.json ]]; then
```

`[[ ]]` isn't POSIX. Eval'd under `dash`/`sh`, it errors and the check silently never runs `fnm use`:

```
sh: [[: not found
```

**Repro**

```
#!/bin/sh
eval "$(fnm env --use-on-cd)"
cd .
```

Run with dash script.sh.

**Fix**

There is a really simple straightforward solution that works everywhere:
```
# instead of:
if [[ -f .node-version || -f .nvmrc || -f package.json ]]; then
# use:
if [ -f .node-version ] || [ -f .nvmrc ] || [ -f package.json ]; then
```

Valid in POSIX `sh` and bash/zsh — safe as the default

RohithPariki:
Fixes #1591 by using the POSIX compliant \[ ]\ syntax instead of the bash-only \[[ ]]\ in the script generated by \nm env --use-on-cd\ for bash and zsh. This ensures compatibility for users whose \/bin/sh\ is \dash\ and who eval the output of \nm env\ in global profile scripts.


RohithPariki:
Closing PR: This was an automated experimental submission while testing a new contribution workflow, and upon review, the implementation is too trivial to justify maintainer review time. Apologies for the noise!

RohithPariki:
Closing PR: Re-evaluated and this is too trivial of a change. Closing permanently.

